PR #697, opened 2015-02-28
fixed #413 email checking for to/cc/cci in webmail

### LINKED ISSUE #413: in the To:, Cc: and Bcc: fields, generate an error when email and/or delimiter is invalid
**Originally created by Fabrice Flore-Thébault on 2013-06-05T13:11:42Z**

In the webmail, if you add multiple emails with a wrong separator, or invalid emails, the application should report that you mistyped something, cancel the sending of the message, and ask you to correct the input.

Additionnaly some advice about the correct formatting of an email / a list of emails could be a nice feature.

vingeni:
fixed #413 email checking for to/cc/cci in webmail

### DIFF
--- a/modoboa/extensions/webmail/templates/webmail/compose.html
+++ b/modoboa/extensions/webmail/templates/webmail/compose.html
@@ -8,7 +8,7 @@
       <div class="form-group{% if form.errors.to %} has-error{% endif %}">
         <label class="col-sm-1 control-label" for="id_to">{% trans "To" %}</label>
         <div class="col-sm-8">
-          <input id="id_to" name="to" class="form-control" value="{% if form.to.value %}{{ form.to.value }}{% endif %}"{% if form.to.errors %} placeholder="{% for error in form.to.errors %}{{ error }}{% endfor %}"{% endif %}>
+          <input id="id_to" name="to" class="form-control" value="{% if form.to.value %}{{ form.to.value }}{% endif %}"{% if form.to.errors %} placeholder="{% for error in form.to.errors %}{{ error }}{% endfor %}" title="{% for error in form.to.errors %}{{ error }}{% endfor %}"{% endif %}>
         </div>
         {% if not form.cc.value %}
         <div class="col-sm-2">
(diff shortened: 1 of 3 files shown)


modoboa/extensions/webmail/validators.py
```diff
@@ -0,0 +1,17 @@
+from django.utils.encoding import force_text
+from django.utils.deconstruct import deconstructible
```
tonioo: Looks like this import is never used.

modoboa/extensions/webmail/validators.py
```diff
@@ -0,0 +1,17 @@
...
+
+class EmailListValidator(object):
+    """
+    Validate a comma separated list of email
+    """
+    def __call__(self, value):
+        value = force_text(value)
+        emails = map(unicode.strip, value.split(','))
+        addresses = getaddresses(emails)
+        [ validate_email(e) for r, e in addresses ]
```
tonioo: I think spaces before and after brackets are not PEP8 compatible.

modoboa/extensions/webmail/templates/webmail/compose.html
```diff
@@ -8,7 +8,7 @@
       <div class="form-group{% if form.errors.to %} has-error{% endif %}">
         <label class="col-sm-1 control-label" for="id_to">{% trans "To" %}</label>
         <div class="col-sm-8">
-          <input id="id_to" name="to" class="form-control" value="{% if form.to.value %}{{ form.to.value }}{% endif %}"{% if form.to.errors %} placeholder="{% for error in form.to.errors %}{{ error }}{% endfor %}"{% endif %}>
+          <input id="id_to" name="to" class="form-control" value="{% if form.to.value %}{{ form.to.value }}{% endif %}"{% if form.to.errors %} placeholder="{% for error in form.to.errors %}{{ error }}{% endfor %}" title="{% for error in form.to.errors %}{{ error }}{% endfor %}"{% endif %}>
```
tonioo: If you include the errors into the title property, the user won't see them unless he puts the mouse on the input box, right ?

vingeni: Yes, I added the error message in the title, because it was set 
previously only in the placeholder field which is not displayed if the 
field is not empty.

We could do something else for the error messaging system, if you have 
an Idea...

--
vincent.

tonioo: Ok. I don't have any idea atm but this is something we can handle later.

PR #699, opened 2015-03-12
Added examples for CSV import

bennidi:
I found the descriptions for the import rather confusing and needed several tries to get it to work. I think examples help.


tonioo:
Indeed, it is a good idea. The problem is that form is becoming fat... Users have to scroll down in order to display everything. I wonder if it would be better to separate this help from the form. What do you think ?

bennidi:
I think that could make sense. It could be moved into a tooltip and shown
onMouseOver or in a collapsible section (I would prefer that). Collapsible
sections should be fairly easy to setup with a bit of jquery and do not
limit the type and amount of data to be displayed.

2015-03-17 10:35 GMT+01:00 Antoine Nguyen <notifications@github.com>:

tonioo:
I merged your PR too quickly, I didn't notice you modified the translation file. You need modify the source template (html file) instead. The translation file will be automatically updated after.

PR #706, opened 2015-04-09
Gather statistics for spam and virus messages

fyfe:
Add support for gathering statistics about spam and virus emails.

### DIFF
--- a/modoboa/extensions/stats/management/commands/logparser.py
+++ b/modoboa/extensions/stats/management/commands/logparser.py
@@ -297,12 +297,13 @@ class LogParser(object):
         if not m:
             return
         host, prog, pid, log = m.groups()
-        m = re.search("(INFECTED|SPAMMY) .* <[^>]+> -> <[^@]+@([^>]+)>.*", log)
+        m = re.search("(INFECTED|SPAM|SPAMMY) .* <[^>]+> -> <[^@]+@([^>]+)>.*",
+                      log)
         if m is not None:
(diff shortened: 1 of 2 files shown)


modoboa/extensions/stats/management/commands/logparser.py
```diff
@@ -297,6 +297,14 @@ def _parse_line(self, line):
         if not m:
             return
         host, prog, pid, log = m.groups()
+        m = re.search("(INFECTED|SPAMMY) .* <[^>]+> -> <[^@]+@([^>]+)>.*", log)
```
tonioo: And what about SPAM ?

tonioo:
That's a great feature, thank you :)

modoboa/extensions/stats/management/commands/logparser.py
fyfe: The regular expression catches the type virus (INFECTED) or spam (SPAMMY) in group 1, group 2 contains the domain. The if statements on lines 303 and 305 identify it its spam or a virus and increment the relevant counter.

tonioo: Sorry, my question was unclear. In addition to the keywords you already use, there is another one which is ```SPAM```. I just wonder why you don't catch it.

fyfe: Because I didn't have any in my logs :)

PR #718, opened 2015-04-29
Improved: mail selection in shift mode.

vingeni:
This is now the same behavior than a famous webmail.


tonioo:
Thank you , it works like a charm!

PR #722, opened 2015-04-30
Fix #717

### LINKED ISSUE #717: In auto-reply message form, some issue around AM/PM datetime picker
In auto-reply message form, some issue around AM/PM datetime picker
capture d ecran 2015-04-29 a 15 18 23

I can't select AM or PM.

Suggest : enable 24h format

harobed:
Fix #717 : switch date time in auto-reply form to 24h format + remove readonly on From and Until fields

### DIFF
--- a/modoboa/extensions/postfix_autoreply/general_callbacks.py
+++ b/modoboa/extensions/postfix_autoreply/general_callbacks.py
@@ -9,8 +9,8 @@ from .models import ARmessage
 def extra_js(user):
     return ["""function autoreply_cb() {
     $('.datefield').datetimepicker({
-        format: 'YYYY-MM-DD hh:mm:ss',
-        language: '%(lang)s'
+        format: 'YYYY-MM-DD HH:mm:ss',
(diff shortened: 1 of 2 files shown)


modoboa/extensions/postfix_autoreply/general_callbacks.py
```diff
@@ -9,8 +9,8 @@
 def extra_js(user):
     return ["""function autoreply_cb() {
     $('.datefield').datetimepicker({
-        format: 'YYYY-MM-DD hh:mm:ss',
-        language: '%(lang)s'
+        format: 'YYYY-MM-DD HH:mm:ss',
+        language: '%(lang)s',
```
tonioo: This extra comma is not supported by all browsers. This is not Python ;)

harobed: Yes, I'll fix it.

harobed:
I removed extra comma.

PR #726, opened 2015-05-04
Fix #713 : now it is possible to create alias with tag (+) in recipient ...

### LINKED ISSUE #713: Create alias with tag (+) in recipient address with internal domain
It's impossible to create new alias with tag in recipient address.

Example : 

* I've contact@example.com mailbox
* I would like to create foobar@example.com alias with contact+test@example.com recipient

I've this error « Local recipient contact+test@example.com not found ».

Solution proposition : 

* use Alias.extmboxes to record this email address with tag

Do you see other solution ?

harobed:
now it is possible to create alias with tag (+) in recipient address with internal domain. See #713

### DIFF
--- a/modoboa/extensions/admin/forms/alias.py
+++ b/modoboa/extensions/admin/forms/alias.py
@@ -96,31 +96,36 @@ class AliasForm(forms.ModelForm, DynamicForm):
                 raise BadRequest(
                     u"%s %s" % (_("Invalid mailbox"), v)
                 )
-            try:
-                domain = Domain.objects.get(name=domname)
-            except Domain.DoesNotExist:
-                domain = None
+
+            # Support tag in recipient, see https://github.com/tonioo/modoboa/issues/713
+            local_part_with_tag = None
+            if '+' in local_part:
+                local_part_with_tag = local_part
+                local_part = local_part[0:local_part.find('+')]
+
+            domain = Domain.objects.filter(name=domname).first()
+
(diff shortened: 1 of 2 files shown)


modoboa/extensions/admin/forms/alias.py
```diff
@@ -96,31 +96,36 @@ def set_recipients(self):
...
-            try:
-                domain = Domain.objects.get(name=domname)
-            except Domain.DoesNotExist:
-                domain = None
+
+            # Support tag in recipient, see https://github.com/tonioo/modoboa/issues/713
+            local_part_with_tag = None
+            if '+' in local_part:
+                local_part_with_tag = local_part
+                local_part = local_part[0:local_part.find('+')]
```
tonioo: Is it going to work if more than one "+" is contained in the local part ?

harobed: Yes it work, only first + is used.

tonioo: Ok fine.

tonioo:
Your PR looks fine but could you please add some test cases into the admin app? There are already test cases around aliases management so it should not be too complicated.

harobed:
> Your PR looks fine but could you please add some test cases into the admin app? There are already test cases around aliases management so it should not be too complicated.

It's done in my last commit.

harobed:
Thanks

PR #728, opened 2015-05-04
Fix remote_addr if modoboa django app is powered by apache or nginx proxy

harobed:
Fix remote_addr if modoboa django app is powered by apache or nginx proxy


harobed:
remote_addr is used here : https://github.com/tonioo/modoboa/blob/master/modoboa/core/views/auth.py#L51

tonioo:
Thanks for the fix!

PR #729, opened 2015-05-05
Fix typo, revise German translation

disko:
Fix typo in help text and all corresponding translations' msgids, also fix translation of that item in German and Dutch.

Revise German translation.


tonioo:
Thank you!

PR #731, opened 2015-05-11
Append imaputils to modoboa main repository

harobed:
Fix https://github.com/modoboa/modoboa-sievefilters/issues/4 append imaputils to modoboa main repository

### DIFF
--- a/modoboa/core/app_settings.py
+++ b/modoboa/core/app_settings.py
@@ -235,6 +235,26 @@ class GeneralParametersForm(parameters.AdminParametersForm):
         widget=forms.Select(attrs={"class": "form-control"})
     )
 
+    sep4 = SeparatorField(label=_("IMAP settings"))
+
+    imap_server = forms.CharField(
+        label=_("Server address"),
+        initial="127.0.0.1",
+        help_text=_("Address of your IMAP server")
+    )
+
+    imap_secured = YesNoField(
+        label=_("Use a secured connection"),
+        initial="no",
+        help_text=_("Use a secured connection to access IMAP server")
+    )
+
+    imap_port = forms.IntegerField(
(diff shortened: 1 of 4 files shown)


harobed:
I've removed useless method in imaputils. Tested on my instanace with sievefilter

tonioo:
Actually, moving those modules into the core application is not a really good idea. Parameters are splitted between applications (core and webmail), this is not consistent... You also need to migrate existing parameters (rename them from modoboa_webmail.XXX to core.XXX)

Other issue: people who don't use the webmail and the sieve module will see useless IMAP parameters. I definitely think this is not the right place.

Maybe a cleaner solution would be to create a dedicated library (modoboa_imap_library?) for that ?

harobed:
> Other issue: people who don't use the webmail and the sieve module will see useless IMAP parameters. I definitely think this is not the right place.

I understand, what is a best solution ? append this imap_utils in modoboa_sievefilters ?

harobed:
> Maybe a cleaner solution would be to create a dedicated library (modoboa_imap_library?) for that ?

If we create too many small libs, I think that modoboa will become a kludge.

tonioo:
Indeed... I don't know what to do right now. If we copy the modules into the sieve extensions, we will have to configure IMAP parameters twice (webmail and sieve).

harobed:
>  If we copy the modules into the sieve extensions, we will have to configure IMAP parameters twice (webmail and sieve).

Now, it's can be a good compromise.

tonioo:
Finally, I think you're right. Users will have to configure IMAP settings twice.
Anyway, we don't have to put the fetch_parser.py file into the core application since it is only used by the webmail. I will split this part.

PR #739, opened 2015-06-10
Before this fix, role is always empty when AccountCreated event was executed

harobed:
Before this fix, role is always empty when AccountCreated event was executed (see https://github.com/modoboa/modoboa-admin-limits/pull/3 )

### DIFF
--- a/modoboa/core/models.py
+++ b/modoboa/core/models.py
@@ -377,8 +377,9 @@ class User(PermissionsMixin):
         self.first_name = row[3].strip()
         self.last_name = row[4].strip()
         self.is_active = (row[5].strip() in ["True", "1", "yes", "y"])
-        self.save(creator=user)
+        self.save()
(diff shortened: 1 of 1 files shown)


modoboa/core/models.py
```diff
@@ -379,6 +379,8 @@ def from_csv(self, user, row, crypt_password=True):
         self.is_active = (row[5].strip() in ["True", "1", "yes", "y"])
         self.save(creator=user)
```
tonioo: The ```creator``` argument can be removed.

modoboa/core/models.py
```diff
@@ -379,6 +379,8 @@ def from_csv(self, user, row, crypt_password=True):
...
         self.set_role(desired_role)
+        self.save()
```
tonioo: This save is useless

Issue #742, opened 2015-06-15
Incorrect "for" attribute on label tags


eberkund:
Something I noticed when using the demo at https://demo.modoboa.org is that all of the HTML `for` attributes for the `<label>` tags are incorrectly suffixed with an addional underscore and number so they do not match the `id` attribute in the preceding `<input>` tag like they are supposed to. This causes clicking on labels to not check the radio box like they are supposed to.

PR #743, opened 2015-06-15
Fix incorrect label tag

eberkund:
The value of the for attribute must match the id of the input tag


tonioo:
Hi,
looking at your PR, I realize the widget you modified is not used anymore... Can you tell me where (which page(s)) you encounter the behaviour you describe in your issue please ?

eberkund:
See here: https://github.com/tonioo/modoboa/issues/742

It occurs when there is multiple radio buttons, here is a screenshot: http://i.imgur.com/UvKSmDZ.png

tonioo:
Ok I see. Looks fine!

PR #749, opened 2015-06-23
Append « in seconds » in « Top notifications check interval » field help message

harobed:
Append « in seconds » in « Top notifications check interval » field help message.


harobed:
This is against 1.3.x branch.

harobed:
Thanks

PR #750, opened 2015-06-23
Fix check new versions bug if modoboa is offline

harobed:
Fix check new versions bug if modoboa is offline (example in train)

### DIFF
--- a/modoboa/core/extensions.py
+++ b/modoboa/core/extensions.py
@@ -118,13 +118,16 @@ class ExtensionsPool(object):
     def list_all(self, check_new_versions=False):
         """List all defined extensions."""
         result = []
+        new_extensions = {}
         if check_new_versions:
             new_extensions = ModoAPIClient().list_extensions()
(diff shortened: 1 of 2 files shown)


modoboa/core/extensions.py
```diff
@@ -118,13 +118,14 @@ def load_all(self):
...
         result = []
+        new_extensions = {}
         if check_new_versions:
-            new_extensions = ModoAPIClient().list_extensions()
-            new_extensions = dict(
-                (ext["name"], ext["version"]) for ext in new_extensions
-            )
-        else:
-            new_extensions = {}
+            tmp = ModoAPIClient().list_extensions()
```
tonioo: Could you rename tmp to something more verbose (like new_extensions as before) ?

harobed: Ok

harobed:
tmp variable removed.

harobed:
Thanks

PR #756, opened 2015-07-04
Fix invalid path for settings.css

carragom:
The path to settings.css seems to be invalid in settings_headers.html template.


tonioo:
Hi Carlos, thanks for the report! Actually, I think I'm going to remove this CSS file so could you just remove the line from the template please ?

PR #794, opened 2015-11-27
First draft of a REST API.

tonioo:
see #523

PR #796, opened 2015-12-01
Correct INSTALLED_APPS in devmode

euank:
PR #794 inadvertently missed a comma that results in, when devmode is set, a dependency of 'authtokendjangobower' attempting to be installed.
This splits that into the two dependencies it was meant to be.


tonioo:
Indeed, thank you for the fix!

Issue #776, opened 2015-11-02
[doc] Missing information about extensions


daks:
Hello,

I installed modoboa several times these days, and I got a lot of problems dealing with extensions. I didn't find sufficient information about which are available, mandatory and more.

I installed in a virtualenv each time, the first one without extensions ('less is more secure' :) but I needed at least modoba_admin.
Another one I installed them all: I got an 'internal error server' and only adding DEBUG=True helped me understand the amavisd extension needed configuration.
Another time, I only wanted to add an extension and can't find how. (I imagine a 'pip install' and adding the app in settings.py do the job, i didn't tried)

Those are examples of problems I dealt with. Maybe the doc is there but I just couldn't find it.
If it's not the case, could it be possible to add some? I can help writing a few if you indicate me how you want to integrate it.

Thanks.

tonioo:
Hi,

it is true that the current documentation would need some work. Since the split between the core and extensions of Modoboa, I realize it was a bit too violent. In a near future, I'll merge back the core with 3 extensions: modoboa-admin, modoboa-admin-relaydomains, modoboa-admin-limits. (but the only required one is the first)

Anyway:
* the way to install an extension is available in the README.rst file of each extension (which is readable from the Github page)
* the "internal server error" you encountered is probably due a bug with the pdfcredentials extension (I have to fix it)
* your help to improve the documentation would be really appreciated.

A pull-request can do the job. We can also talk about the content and the way to organize it before you start. Feel free to contact me on my personal address :)

tonioo:
@daks What do we do with this issue ?

daks:
I saw you started to revert back extensions to the standard installation, if this work is finished and the docs on RTD is ok, I can look at it again and propose some doc enhancement.
I'll try to take some time next week.

tonioo:
@daks yes, this work is done.

PR #813, opened 2015-12-09
Plugins doc

daks:
Hello,

I finally made very few changes on the doc to add information about plugins. Each one having its own doc, I only linked to the index page which contains the list of them.
Ref #776


tonioo:
Makes sense, thank you!

PR #842, opened 2016-03-07
run tests with mysql

gawel:
work in progress. need a PR to run travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,22 +3,33 @@ language: python
 python:
   - "2.7"
 
+env:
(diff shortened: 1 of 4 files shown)


gawel:
It works. But got 3 failures with mysql.

tonioo:
Thanks, I'll look at it.

gawel:
I've fixed an easy one. The others looks more tricky.

gawel:
(and I guess the export command is not really tested since it should fail ;p)

PR #850, opened 2016-03-16
initial implementation of BCRYPTHasher with unit test

cl0secall:
based on tag 1.4.5. adds BCRYPTHasher to modoboa.core.password_hashers.advanced, along with necessary glue. This implementation fixes "rounds" at 12.

### DIFF
--- a/modoboa/admin/tests/password_schemes.py
+++ b/modoboa/admin/tests/password_schemes.py
@@ -34,6 +34,9 @@ class PasswordSchemesTestCase(ModoTestCase):
         self.assertTrue(account.password.startswith(startpattern))
         self.assertTrue(account.check_password('Toto1234'))
 
+    def test_bcrypt_scheme(self):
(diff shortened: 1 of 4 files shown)


tonioo:
Thank you! To let users define a round level different than 12, we need to find a way to limit the value per hasher. I think it can be done later. Have you check what happen when you try to use this hasher on a Linux platform ?
Otherwise, could you please update the corresponding unit test (modoboa/admin/tests/test_password_schemes.py) and so increase the coverage level of your PR ?

tonioo:
@cl0secall Any news about the coverage? :)

cl0secall:
unit test added, plus naming fixes so the test actually passes.

tonioo:
@cl0secall Great! Unfortunately, I made changes to master so now your branch can't be merged automatically :(
I'll try to update it but I don't know if I can easily.

PR #869, opened 2016-05-02
Fix swagger app name in INSTALLED_APPS

cubicuboctahedron:
Hi,
there is a typo in the swagger's module name - `.` instead of `_`.


tonioo:
Thank you !

PR #912, opened 2016-09-13
allow to use tox to run tests/serve locally in an easy way

gawel:
This allow to use tox locally with sqlite (no ldap/postgres/complex stuff). This may ease contributions!

The only thing missing is a few fixtures so you can login when using tox -e serve. Any idea how to do that ? I guess you already have some code for that at least for testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .*~$
 *.pyc
 *.mo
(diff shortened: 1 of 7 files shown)


tonioo:
Thank you !

PR #934, opened 2016-10-06
localization of email headers, added the datetime format settings in …

christianwgd:
…test_project settings.py with some docs

Localization of the email header fields by adding a template filter to lib_tags.py that translates the header labels "From", "To", "Date" and "Subject" and is used in the mailheaders template. Added the required translation strings to the locale po files.


test_project/test_project/settings.py
```diff
@@ -106,6 +106,25 @@
...
+# Webmail datetime formats according to
+# https://docs.djangoproject.com/en/1.10/ref/templates/builtins/
+# modoboa-webmail is able to show the datetime of emails
+# in two ways.
+# If none is provided, default is RFC 5322 formatted date.
+# For this, USE_L10N must be set to True!
+
+# emails of the last week
+# en-us style
+WEBMAIL_SHORT_DATETIME_FORMAT = "l, P"
```
tonioo: I know part of the code used by the webmail might not be at the right place but the settings related to the webmail should be move the webmail repository. You can take example on the amavis plugin: https://github.com/modoboa/modoboa-amavis/blob/master/modoboa_amavis/settings.py.

test_project/test_project/settings.py
```diff
@@ -106,6 +106,25 @@
...
+# european style
+# WEBMAIL_SHORT_DATETIME_FORMAT = "l, H:i"
+
+# emails older than a week
+# en-us style
+WEBMAIL_DATETIME_FORMAT = "N, j Y P"
+# european style
+# WEBMAIL_DATETIME_FORMAT = "d. N Y H:i"
```
tonioo: Same here.

modoboa/lib/templatetags/lib_tags.py
```diff
@@ -94,3 +95,12 @@ def extra_static_content(caller, st_type, user):
...
     )
+
+
+@register.filter(name='localize_header_name')
+def localize_header_name(headername):
+    """ Localizes the header names """
+    names = {'From': _('From'), 'To': _('To'), 'Date': _('Date'), 'Subject': _('Subject')}
+    if names.has_key(headername):
+        return names[headername]
+    return headername
```
tonioo: You can simplify those 3 lines into 1 like this:
```python
return names.get(headername, headername)
```

modoboa/lib/templatetags/lib_tags.py
```diff
@@ -94,3 +95,12 @@ def extra_static_content(caller, st_type, user):
                 "GetStaticContent", caller, st_type, user)
         })
     )
+
+
+@register.filter(name='localize_header_name')
+def localize_header_name(headername):
+    """ Localizes the header names """
+    names = {'From': _('From'), 'To': _('To'), 'Date': _('Date'), 'Subject': _('Subject')}
```
tonioo: Could you please use " instead of ' for strings. I'm currently updating the code to be consistent but it's a long path...

christianwgd:
I put the settings into modoboa because i didn’t find one in modoboa-webmail. Thanks to your hint, i will create one. No problem with the string delimiters, i will change them.

PR #946, opened 2016-10-16
Some corrections of German translation (and translations in general)

christianwgd:
Improved German translation: Translation of user settings dialog,  corrected wrong parenthesis in logging of login/logout, language label field had no label tag


modoboa/locale/de/LC_MESSAGES/django.po
```diff
@@ -426,23 +420,25 @@ msgstr "Passwort"
 msgid "last login"
 msgstr "Letzte Anmeldung"
 
+#: core/models.py:70
+msgid "language"
+msgstr "Sprache"
```
tonioo: Looks like you did not update the source file. Have you run the following command:
```python
> django-admin makemessages -a
```
?

modoboa/core/models.py
```diff
@@ -67,7 +67,7 @@ class User(PermissionsMixin):
         ugettext_lazy('last login'), blank=True, null=True
     )
 
-    language = models.CharField(
+    language = models.CharField(ugettext_lazy('language'),
```
tonioo: Double quotes please.

PR #972, opened 2016-11-23
avoid failure when using tox/sqlite

gawel:


tonioo:
I fixed this issue: https://github.com/tonioo/modoboa/pull/973.

PR #1006, opened 2017-01-17
Allow local addresses for forward definition.

### LINKED ISSUE #998: forward : Permission denied
hi

forwarding an account to a local destination isn't working, even when set as admin:
Permission denied: You can't define a forward to a local destination. Please ask your administrator to create an alias instead.

maybe i miss something?
there is
- account1@foo.bar
- account2@foo.bar

during holyday of account1 the emails should be forwarded to account2@foo.bar (with/without keeping local copy as account1 likes).
how to solve this?

thanks a lot

tonioo:
fix #998

### DIFF
--- a/modoboa/admin/forms/forward.py
+++ b/modoboa/admin/forms/forward.py
@@ -1,15 +1,13 @@
 """Forms related to forwards management."""
 
 from django import forms
-from django.utils.translation import ugettext as _, ugettext_lazy
-
-from modoboa.lib.email_utils import split_mailbox
-from modoboa.lib.exceptions import BadRequest, PermDeniedException
-
-from ..models import Domain
+from django.core import validators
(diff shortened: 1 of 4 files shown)


cmdtstof:
means implemente?
thanks
stof

tonioo:
@cmdtstof yes, but only in the master branch. I will be released with the next modoboa version (1.7.0)

cmdtstof:
then another €100 is on your way
and we try to start a
!!! crowdfunding for a modoboa auto update function !!!
thanks

tonioo:
@cmdtstof Thank you!
Are you joking about the crowdfunding campaign ?

cmdtstof:
@tonioo
not really. there is already a positive reaction for it! :-)
let's think about it via PM

PR #1029, opened 2017-02-13
Dovecot MySQL queries update

yannfill:
Using the MySQL password query for Dovecot found in Modoboa documentation, the user is still able to authenticate when his domain (and not his email) is disabled. Updated query prevents that.

Also, iterate query should select email as user instead of username since doveadm command applies to whole email address.

And cosmetic changes for user query.

### DIFF
--- a/doc/manual_installation/dovecot.rst
+++ b/doc/manual_installation/dovecot.rst
@@ -147,11 +147,11 @@ MySQL users
 
   default_pass_scheme = CRYPT
 
-  password_query = SELECT email AS user, password FROM core_user WHERE email='%u' and is_active=1
+  password_query = SELECT email AS user, password FROM core_user, admin_domain WHERE ecore_user.mail='%Lu' AND core_user.is_active=1 AND admin_domain.name='%Ld' AND admin_domain.enabled=1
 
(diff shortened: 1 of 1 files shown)


doc/manual_installation/dovecot.rst
```diff
@@ -147,11 +147,11 @@ MySQL users
 
   default_pass_scheme = CRYPT
 
-  password_query = SELECT email AS user, password FROM core_user WHERE email='%u' and is_active=1
+  password_query = SELECT email AS user, password FROM core_user, admin_domain WHERE core_user.email='%Lu' AND core_user.is_active=1 AND admin_domain.name='%Ld' AND admin_domain.enabled=1
```
tonioo: Since there is no join in your query, have you checked performance?

yannfill: Not at all, I'm not good at DBA. I'm running a server with about 150-200 simultaneous connections ans it seems fine.

Talking about optimization, maybe you could modify the password query so that user query is done at the same time. See http://wiki2.dovecot.org/UserDatabase/Prefetch

tonioo:
Actually, I realize the documentation is wrong. I made some changes to the installer (https://github.com/modoboa/modoboa-installer/pull/104) including prefetching. Now I'm going to update the documentation.

tonioo:
@yannfill Please take a look at #1039 and thank you for reporting.

PR #1044, opened 2017-02-21
Notes about quota changes and Reseller user

pavinjosdev:
Notes about quota changes and Reseller user in v1.7.0.

Current behavior before PR:


tonioo:
I think this text should be included to the upgrade.rst file, for 1.7.0 changes. What do you think?

pavinjosdev:
@tonioo Yes, that would be better :)

PR #1122, opened 2017-05-10
make modoboa-admin deploy work with python 3.4

arthru:
The command `modoboa-admin.py deploy --collectstatic --dburl default:postgres://localhost/modoboa_test --domain localhost --admin-username admin modoboa_test` failed with a lot of errrors when launched with python 3.4. The PR aims to  fix that.

Current behavior before PR:
Errors when launching ; return code != 0

No errors ; return code == 0

### DIFF
--- a/modoboa/admin/management/commands/subcommands/_repair.py
+++ b/modoboa/admin/management/commands/subcommands/_repair.py
@@ -4,6 +4,7 @@ from __future__ import print_function, unicode_literals
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
+from django.utils.encoding import smart_str
 
 from modoboa.lib.permissions import grant_access_to_object
 from modoboa.lib.permissions import get_object_owner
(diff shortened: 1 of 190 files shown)


modoboa/admin/management/commands/subcommands/_repair.py
```diff
@@ -4,6 +4,7 @@
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
+from django.utils.encoding import smart_str
```
tonioo: smart_text ?

arthru: smart_str returns str in both python 2 and 3 whereas smart_text returns unicode in python 2 and str in python3. Sometimes, we need smart_str when the API keeps using str in both python 2 and 3....

arthru:
So, the tests are ok in both python 2 and python 3...

I am going to install a live instance on this branch to validate that everythings work fine.

arthru:
BTW, I don't know why there is a problem with codecov/project... and i can't access the "Details" page...

PR #1140, opened 2017-06-17
Remove maybe missleading prefix

frlan:
At this very point you are not yet inside a virtualenv.


tonioo:
@frlan Indeed! Thank you.

PR #1185, opened 2017-07-13
docs: Mention WSGIPassAuthorization in Apache config

a3li:
This is needed to have the Authorization header passed to the API.
Took me a while to figure this out, so it would probably be good to document it.


tonioo:
@a3li Thank you.

PR #1186, opened 2017-07-14
Fix redirect url

floreal797:
When we logout, the new url has not the "next" get parameter. nextlocation is not None in this case, but "None" (the string). 

Current behavior before PR:
So if we re-login, we are redirected to accounts/login/None, that does not exists (404).

I add ' != "None" ' to the conditions, so now the re-login works.


tonioo:
@floreal797 Thank you.

PR #1192, opened 2017-07-24
Python3 compatibility: add __str__ functions in addition to __unicode__ function

floreal797:
Python 3 doesn't interpret __unicode__ function, but use __str__ function instead.

Current behavior before PR:
So, for example, in DNSBL summary, instead of the domain name and IP in the second column, we have "MXRecord object"...

Restore the view of domain name and IP, like python2 version.

I also add a __str__ function beside every __unicode__ function. Note that python3 uses unicode by default, so I remove in str function the "u" prefix for strings.

The change in file modoboa/admin/models/domain.py fixes the problem described above. The other changes have not been tested by my. Please check they are correct.

### DIFF
--- a/modoboa/admin/models/domain.py
+++ b/modoboa/admin/models/domain.py
@@ -315,6 +315,7 @@ class MXQuerySet(models.QuerySet):
         return self.exists()
 
 
+@python_2_unicode_compatible
 class MXRecord(models.Model):
     """A model used to store MX records for Domain."""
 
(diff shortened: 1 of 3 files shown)


tonioo:
Thank you. Could you also use ``python_2_unicode_compatible`` decorator? (https://docs.djangoproject.com/fr/1.10/ref/utils/#django.utils.encoding.python_2_unicode_compatible)

floreal797:
As described in the doc, I removed `__unicode__` functions (already replaced by `__str__` functions), and add the `python_2_unicode_compatible` decorator to the parent class. Is it correct?

tonioo:
@floreal797 Yes, it is. Thank you.

PR #1257, opened 2017-09-29
Solve issues with deploy assistant

alexander255:

  1. Deploy assistant uses `eval(input(…))` when showing prompts
  2. Deploy assistant invokes `manage.py` using `python`(2) instead of the current interpreter

Current behavior before PR:

ad (1): `UnboundLocalError`, all entries have to be quoted by the user or exceptions or worse may happen (and on Python 2 everything is `eval`'d twice!!)
ad (2): Deploy fails if system Python is Python 2, but one tries to deply Modoboa using Python 3

No more errors on deploy.


tonioo:
Thank you!

Issue #1318, opened 2017-12-07
pl_PL translation


tonioo:
Add the new pl_PL translation.

PR #1319, opened 2017-12-07
Added polish translation.

tonioo:
see #1318

### DIFF
--- a/modoboa/core/constants.py
+++ b/modoboa/core/constants.py
@@ -36,6 +36,7 @@ LANGUAGES = (
     ("nl", u"nederlands"),
     ("pt_PT", u"português"),
     ("pt_BR", u"português (BR)"),
+    ("pl_PL", u"polski"),
     ("ru", u"русский"),
     ("sv", u"svenska"),
 )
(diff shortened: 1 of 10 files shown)


modoboa/core/constants.py
```diff
@@ -36,6 +36,7 @@
     ("nl", u"nederlands"),
     ("pt_PT", u"português"),
     ("pt_BR", u"português (BR)"),
+    ("pl_PL", u"polski"),
     ("ru", u"русский"),
```
xiejing178: Hello,why am I added Chinese here, but there is no Chinese option in Settings

tonioo: Which code do you use? Are you running the development server?

xiejing178: In fact,I don't know the answer,because I just came to this company for few days and my boss asked me to do this.So,maybe I should change the question,when will you have Chinese?

tonioo: When someone will translate Modoboa to Chinese.

xiejing178: okay ! thank you .

PR #1330, opened 2017-12-17
Improved email parser

fyfe:
This is a refactor of `moboboa.lib.email_utils.Email`, currently it has some dead code (`__save_image`) and it has trouble parsing messages containing certain attachemnt types.

73bceee splits `modoboa/lib/tests.py` into separate files so that each file only tests one component. The code in `modoboa/lib/tests/__init__.py` is there so as not to break everything that imports `modoboa.lib.tests`

ce02d24 adds python 3 support for landscape.io and simplifies codecov ignore paths

c67f685 and e38cfb7 adds the improved `Email` class and tests.

**This PR depends on a similar PR for modoboa/modoboa-amavis#80 as it depends on moboboa.lib.email_utils.Email.**

modoboa-webmail isn't affected as it implements it's own parser to fetch e-mails from IMAP.

### DIFF
--- a/modoboa/lib/email_utils.py
+++ b/modoboa/lib/email_utils.py
@@ -1,21 +1,28 @@
-# coding: utf-8
+# -*- coding: utf-8 -*-
 
 from __future__ import unicode_literals
 
-from email.header import Header, decode_header
+import email
+from email.header import Header
 from email.mime.text import MIMEText
 from email.utils import make_msgid, formatdate, parseaddr
 import re
 import smtplib
 import time
 
-from django.conf import settings
-from django.utils.encoding import smart_text
-from django.utils.html import conditional_escape
+from django.utils.encoding import smart_str, smart_text
+from django.utils.html import conditional_escape, escape
 from django.template.loader import render_to_string
-from django.utils import six
 
 from modoboa.lib import u2u_decode
+
+import lxml.html
+from lxml.html.clean import Cleaner
+
+
+# used by Email()
+_RE_REMOVE_EXTRA_WHITESPACE = re.compile(r"\n\s*\n")
+_RE_CID = re.compile(r".*[ ]*cid=\"([^\"]*)\".*", re.I)
 
 
 class EmailAddress(object):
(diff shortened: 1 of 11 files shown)


tonioo:
@fyfe Can you please explain why you consider __parse_image and __save_image methods as dead code?

fyfe:
@tonioo I've reworded the commit
>For html messages inline images are now base64 encoded and embedded into
the  tag rather than being saved to disk.

I had originally removed support for dealing with images, but instead changed it to base64 encode and embed inline images.

modoboa/lib/email_utils.py
```diff
@@ -33,107 +40,102 @@ def __str__(self):
...
+
         self._msg = None
+        self._headers = None
         self._body = None
+        self._images = {}
+
+        # used by modoboa_webmail.lib.imapemail.ImapEmail
+        self.attachments = {}
+        # not used??
+        self.attached_map = {}
```
tonioo: Looks like it is not used anymore...

modoboa/lib/email_utils.py
```diff
@@ -146,85 +148,124 @@ def _parse(self, msg, level=None):
...
+        for part in msg.walk():
+            content_type = part.get_content_maintype()
+            if content_type == "text":
+                self._parse_text(part, level=level)
+
+            else:
+                # I'm  a dumb mail parser and treat all non-text parts as attachments
+                self._parse_attachment(part, level=level)
+
+    def _parse_attachment(self, msg, level=0):
```
tonioo: This method is wrongly named since it only deals with images. Talking about that, looks like you don't store attachments anymore (in ``self.attachments``), right?

modoboa/lib/email_utils.py
fyfe: Oops forgot to remove that.

modoboa/lib/email_utils.py
fyfe: The old code didn't do anything with attachments either it just saved the inline images to disk to be used when the message was displayed.

tonioo: Yes, I checked that yesterday. Attachments were not used since the webmail has its own way to retrieve them. So, I guess we can rename this method to _parse_image or _parse_inline_image?

fyfe:
@tonioo updated with changes

- unused `attached_map` removed
- `_parse_attachment` renamed to `_parse_inline_image`

PR #1333, opened 2017-12-19
Upgrade Django (Django-1.10 no longer supported)

fyfe:
Django 1.10 is now [no longer supported](https://www.djangoproject.com/download/#supported-versions), this PR upgrades modoboa to Django 1.11 (the last version of Django to support Python 2). There's an associated PR for each of the extensions to upgrade them as well.

I've split this into several patches to make it easier to follow:

- e6befb3: make all the travis.yml settings the same across modoboa projects. Adds testing against Python 3.5 and enables *most* Python warnings, the warnings don't affect the test results but they do give you a heads up about things that are depreciated or wrong.

- 3308c56: [PEP479](https://www.python.org/dev/peps/pep-0479/#consequences-for-existing-code) depreciated StopIterators, they should be replace with return.

- 892160d: fixes to support Django 1.11 (and 2.0)
    - MIDDLEWARE_CLASSES is replaced by MIDDLEWARE

    - add MiddlewareMixin to modoboa middlewares

    - on_delete will become a required parameter for ForeignKey() set it to the current default CASCADE

    - on_delete will become a required parameter for OneToOneField() set it to the current default CASCADE

    - reverse and reverse_lazy have moved to django.urls

    - RadioChoiceInput widget has been removed, replace with HorizontalRadioSelect

- e8eb9a8 0cd444b 2e18acb ba86196: upgrade dependencies and fix any errors/warnings.

- 8fb3e24: upgrade Django, the minimum version is pinned to 1.11.8 as this is the latest version that I've done the upgrade with.

- 4fe51d6: random code cleanups.
    - removed duplicate imports

    - whitespace fixes

    - removed unnecessary u"" prefixes

    - added missing r"" prefixes where required

    - added missing default values to DynamicForm() to prevent possible unassigned variable access

- 33b37ec: updated settings.py for new deployments
    - updated Django documentation urls to point to the current version

    - renamed MIDDLEWARE_CLASSES to MIDDLEWARE, MIDDLEWARE_CLASSES is depreciated in Django >= 1.10

    - whitespace and #noqa fixes

- 92963a0: updated test_project, synced with a fresh deployment using Django 1.11.8 and updated settings.py so the tests environment matches a fresh deployment.

---

There is one remaining warning for **Django 2.0** support in urls.py ([example](https://github.com/modoboa/modoboa/blob/master/modoboa/urls.py#L28)):

`RemovedInDjango20Warning: Specifying a namespace in django.conf.urls.include() without providing an app_name is deprecated. Set the app_name attribute in the included module, or pass a 2-tuple containing the list of patterns and app_name instead.`

The simple fix from the description is to add `app_name = "modoboa"` to urls.py however this then requires **every** (and there's a lot of them) `reverse("<url>")` or `reverse("<namespace>:<url>")` to be prefixed with `<app_name>`. The documentation is very thin for this feature so I'm not sure what's the best solution, it's not a real issue until you upgrade to Django 2.0.

---

Extension PRs

- modoboa/modoboa-amavis#81
- modoboa/modoboa-contacts#11
- modoboa/modoboa-dmarc#28
- modoboa/modoboa-imap-migration#29
- modoboa/modoboa-pdfcredentials#25
- modoboa/modoboa-postfix-autoreply#52
- modoboa/modoboa-radicale#17
- modoboa/modoboa-sievefilters#25
- modoboa/modoboa-stats#36
- modoboa/modoboa-webmail#136

**When this is merged extensions will need their requirements.txt updated to point at this version as a minimum for modobo.**


fyfe:
The only change required for a user to upgrade is to rename `MIDDLEWARE_CLASSES` to `MIDDLEWARE` in `settings.py`, so the upgrade documentation needs something along the lines of

> Edit your :file:`settings.py` file and rename `MIDDLEWARE_CLASSES` to `MIDDLEWARE`

.travis.yml
```diff
@@ -1,41 +1,48 @@
-sudo: false
 language: python
-cache: pip
 python:
   - "2.7"
   - "3.4"
+  - "3.5"
```
tonioo: Is it required since we already check against python 3.6?

tonioo:
@fyfe You rock! That's something I had on my TODO list, what a surprise!! How can I contact you directly?

.travis.yml
fyfe: Probably not while you still support Python 2, but I have been caught out by new ~~bugs~~ features during the evolution of Python 3.

fyfe:
@tonioo you're welcome, I've spent the last couple of days snowed in with very little to do 😃 

You can e-mail me at andrew@neptune-one.net.

PR #1336, opened 2017-12-21
On error email admins

fyfe:
Set ADMINS to a list of all the people who get code error notifications. When DEBUG=False Django emails these people the details of exceptions raised in the request/response cycle.

modoboa disables this by overriding `settings.LOGGING`, e5a05f2 restores this feature it's only enabled if `settings.ADMINS` is defined.

51566ce adds utf-8 support when generating `settings.py`, I might have accidentally thrown a utf-8 character into `settings.py.tpl` by accident (now removed) 😃

### DIFF
--- a/modoboa/core/commands/deploy.py
+++ b/modoboa/core/commands/deploy.py
@@ -4,6 +4,7 @@
 
 from __future__ import unicode_literals, print_function
 
+import codecs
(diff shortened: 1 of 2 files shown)


tonioo:
Thank you.

PR #1339, opened 2018-01-03
Add IPv6 support to mx checks

fyfe:
- add support for looking up IPv6 addresses (socket.getaddrinfo)
- add support for International Domain Names in case the DNS reponse contains one
- when looking up an ip address, don't stop on the first failure check all addresses
- add logging for potential configuration errors:
    - DNS configuration errors (i.e. no MX records)
    - server name resolution errors (i.e. resolv.conf doesn't contain a working name server)
    - invalid values in MX records
- update mock results for tests

### DIFF
--- a/modoboa/admin/lib.py
+++ b/modoboa/admin/lib.py
@@ -1,4 +1,4 @@
-# coding: utf-8
+# -*- coding: utf-8 -*-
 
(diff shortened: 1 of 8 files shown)


tonioo:
Awesome, thank you! 👍

PR #1340, opened 2018-01-03
 Fix test database setup

fyfe:
Django automatically renames test databases to 'test_%s' % settings.DATABASES[*]["NAME"], so modoboa_test became test_modoboa_test. There's no point creating the database in `before script` as the test suite creates it.

https://docs.djangoproject.com/en/1.11/topics/testing/advanced/#django.db.connection.creation.create_test_db

- target oldest currently supported database servers, travis currently defaults to PostgreSQL 9.2 which isn't supported anymore.

- move database dependencies out of travis.yml into test_requirements.txt and pin to minimum versions recommended by django.
  See https://docs.djangoproject.com/en/1.11/ref/databases/

- use MySQL root user for tests, we already use the root user (postgres) for PostgreSQL.

- fix typo in modoboa/test_settings for DB variable, default should be POSTGRESQL not POSTGRES

- enable MySQL strict mode (on by default on MySQL >= 5.7)
  See https://docs.djangoproject.com/en/1.11/ref/databases/#mysql-sql-mode

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ env:
 sudo: false
 
 addons:
+  postgresql: "9.3"
+  mysql: "5.5"
   apt:
     packages:
       - ldap-utils
@@ -29,8 +31,6 @@ services:
 
 before_install:
   - pip install codecov
-  - if [[ $DB = "POSTGRESQL" ]]; then pip install -q psycopg2; fi
-  - if [[ $DB = "MYSQL" ]]; then pip install -q mysqlclient; fi
 
 install:
   - pip install -q -r requirements.txt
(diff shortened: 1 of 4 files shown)


tonioo:
Great, thank you :)

tonioo:
But I wonder why codecov check has not been triggered for this PR. Maybe the codecov.yml file rename?

fyfe:
I've undo the rename there's a couple of reports about codecov not liking it being renamed.

There's something wrong with travis/codecov, the build reports for this PR (build [1940](https://travis-ci.org/modoboa/modoboa/builds/325037114?utm_source=github_status&utm_medium=notification))  and #1342 (build [1942](https://travis-ci.org/modoboa/modoboa/builds/325062818?utm_source=github_status&utm_medium=notification)) have been sent to the [last commit on master](https://codecov.io/gh/modoboa/modoboa/commit/f9341633c69a29bd7e31a0e543b9e8576f578372/build) so codecov isn't posting any reports on the PRs.

tonioo:
@fyfe yes I've noticed this too and I don't see any opened issue on codecov side.

PR #1342, opened 2018-01-03
Fix importing csv files containing utf8 characters

### LINKED ISSUE #1312: Error 500 Import CSV
# Impacted versions

* Modoboa: 1.9.1
* installer used: Yes/No ?
Yes
* Webserver: Nginx/Apache
Nginx

# Steps to reproduce
Identities / Import

# Current behavior
I can't import csv with users, Javascript browser console throws 500 Error(see Screenshot) when it's try to access to https://server.com/admin/identities/import/

https://imgur.com/a/CnU7g

fyfe:
The csv module on Python 2 doesn't support utf8 (even encoded as bytes), the easiest fix is to use the csv module backported from Python 3.

I've also replaced `fct = getattr(lib, "import_%s" % row[0].strip())` with a lookup dict `IMPORT_OBJECT_TYPES`. There's a potential here for somebody  to attempt to call other functions with a csv import file.

Fixes #1312

See https://docs.python.org/2/library/csv.html
See https://github.com/ryanhiebert/backports.csv

### DIFF
--- a/modoboa/admin/management/commands/subcommands/_import.py
+++ b/modoboa/admin/management/commands/subcommands/_import.py
@@ -21,7 +21,7 @@ class ImportCommand(BaseCommand):
     """Command class."""
 
     args = "csvfile"
-    help = "Import identities from a csv file"
+    help = "Import identities from a csv file"  # noqa:A003
 
     def add_arguments(self, parser):
         """Add extra arguments to command."""
--- a/modoboa/admin/management/commands/subcommands/_import.py
+++ b/modoboa/admin/management/commands/subcommands/_import.py
@@ -2,7 +2,8 @@
 
 from __future__ import unicode_literals
 
-import csv
+from backports import csv
+import io
 import os
 
 import progressbar
@@ -13,7 +14,6 @@ from modoboa.core import models as core_models
 from modoboa.core.extensions import exts_pool
 from modoboa.lib.exceptions import Conflict
 
-from .... import lib
 from .... import signals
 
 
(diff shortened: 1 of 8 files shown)


modoboa/admin/management/commands/subcommands/_import.py
```diff
@@ -52,20 +53,23 @@ def _import(self, filename, options):
...
-        with open(filename) as f:
+        with io.open(filename, encoding="utf8") as f:
             reader = csv.reader(f, delimiter=options["sepchar"])
             i = 0
             for row in reader:
                 if not row:
                     continue
-                try:
-                    fct = getattr(lib, "import_%s" % row[0].strip())
-                except AttributeError:
```
tonioo: Can you explain why this way of looking for import function was wrong please?

fyfe: `row[0]` is taken directly from the csv file with no checks on what it contains, there's a potential for code injection, I'm not sure how vulnerable python is to this.

Also Python 2 didn't like it when I change domain to d**ó**main in one of the csv file import tests because Python 2 doesn't allow unicode characters in method names.

The other option is to register the standard imports with `admin_signals.import_object` in `admin/handlers.py` the same way as extensions do.
```python
@receiver(admin_signals.import_object)
def get_import_func(sender, objtype, **kwargs):
    """Return function used to import objtype."""
    if objtype == "domain":
        return lib.import_domain
    elif objtype == "domainalias":
        return lib.import_domainalias
    elif objtype == "account":
        return lib.import_account
    elif objtype == "alias":
        return lib.import_alias
    elif objtype == "forward":
        return lib.import_forward
    elif objtype == "dlist":
        return lib.import_dlist
    return None
```
And change this:
https://github.com/modoboa/modoboa/blob/2c4fefca02d1591a847de835b2974e7f6874a509/modoboa/admin/management/commands/subcommands/_import.py#L61-L68
to:
```python
fct = signals.import_object.send(
    sender=self.__class__, objtype=row[0].strip())
if not fct or fct[0][1] is None:
    continue
fct = fct[0][1]
```

tonioo: I don't think python is vulnerable to this but I'm not sure... Anyway, I do prefer this new approach :)

fyfe: fixed

Issue #1325, opened 2017-12-08
Per-domain DKIM key


tonioo:
Add the possibility to generate a DKIM key pair for every managed domain. The public key must be displayed in the domain detail page.

This feature is looking for sponsors, more information [here](https://modoboa.org/sponsor/5/)

Toniob:
Some possible features :
- possibility to select the size of the key
- a task to export the dns record to files (to simply include those files in bind for example)

mroethke:
Just out of curiosity: why opendkim? amavis can do dkim signing as well and most, if not all, modoboa users use it already.

tonioo:
@mroethke Because amavis can't read its configuration from a SQL database...

PR #1352, opened 2018-01-12
Per domain DKIM key generation.

tonioo:
see #1325


modoboa/admin/app_settings.py
```diff
@@ -173,6 +194,22 @@ def clean_default_mailbox_quota(self):
...
 
+    def clean_dkim_keys_storage_dir(self):
+        """Check that directory exists."""
+        storage_dir = self.cleaned_data.get("dkim_keys_storage_dir", "")
+        if storage_dir:
+            if not os.path.isdir(storage_dir):
+                raise forms.ValidationError(
+                    ugettext_lazy("Directory not found")
+                )
+            code, output = exec_cmd("which openssl")
```
fyfe: A pure python solution? `if not which("openssl"):`

```python
import os

def which(program):
    def is_exe(fpath):
        return os.path.isfile(fpath) and os.access(fpath, os.X_OK)

    fpath, fname = os.path.split(program)
    if fpath:
        if is_exe(program):
            return program
    else:
        for path in os.environ["PATH"].split(os.pathsep):
            exe_file = os.path.join(path, program)
            if is_exe(exe_file):
                return exe_file

    return None
```

tonioo: Yes, why not. But it would require to update every calls to ``which`` (not necessarily related to this PR) so I suggest we make a new PR. What do you think?

fyfe: Sure, I'll knock something up.

modoboa/admin/management/commands/subcommands/_manage_dkim_keys.py
```diff
@@ -33,7 +33,7 @@ def create_new_dkim_key(self, domain):
         if code:
             print("Failed to generate DKIM public key for domain {}: {}"
                   .format(domain.name, output))
-        public_key = ""
+        public_key = b""
         for cpt, line in enumerate(output.splitlines()):
```
fyfe: It would be easier to convert output to text
`for cpt, line in enumerate(smart_text(output).splitlines()):`

modoboa/admin/models/domain.py
```diff
@@ -58,6 +58,21 @@ class Domain(AdminObject):
...
         "transport.Transport", null=True, on_delete=models.SET_NULL)
 
+    enable_dkim = models.BooleanField(
+        ugettext_lazy("Enable DKIM signing"),
+        help_text=ugettext_lazy(
+            "If you activate this feature, a DKIM key will be "
+            "generated for this domain."),
+        default=False
+    )
+    dkim_key_selector = models.CharField(max_length=30, default="default")
```
kryskool: please use modoboa as selector to prevent conflict with other mail/mailing apps

modoboa/admin/management/commands/subcommands/_manage_dkim_keys.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def create_new_dkim_key(self, domain):
+        """Create a new DKIM key."""
+        storage_dir = param_tools.get_global_parameter("dkim_keys_storage_dir")
+        pkey_path = os.path.join(storage_dir, "{}.pem".format(domain.name))
+        key_size = (
+            domain.dkim_key_length if domain.dkim_key_length
+            else self.default_key_length)
+        code, output = sysutils.exec_cmd(
+            "openssl genrsa -out {} {}".format(pkey_path, key_size))
```
kryskool: The private key have a chmod 600, the owner must be the user that launch opendkim

modoboa/admin/constants.py
```diff
@@ -99,3 +99,9 @@
     ("domain", _("Domain")),
     ("relaydomain", _("Relay domain")),
 ]
+
+DKIM_KEY_LENGTHS = [
+    (1024, "1024"),
```
kryskool: 1024 key must be insecure https://gsuiteupdates.googleblog.com/2016/05/getting-rid-of-spoofers-digitally-sign.html

fyfe: I disagree, 1024 bit is **less** secure not **in**secure.  1024 bit should still be available (not the default) as there are some DNS providers that still don't support >1024 bit keys, yes it's 2018 but progress is slow 😞

modoboa/admin/management/commands/subcommands/_manage_dkim_keys.py
fyfe: Does openssl not automatically make the private key chmod 600 when it's generated?

modoboa/admin/constants.py
tonioo: Good point. So I'll leave it for now.

modoboa/admin/management/commands/subcommands/_manage_dkim_keys.py
tonioo: Yes, it does. We just need to document that keys will belong to the user running the management command.

kryskool: Yep and check if correct with the installer

PR #1353, opened 2018-01-12
refactor setup.py

fyfe:
Refactor setup.py using latest version of [Pythons packaging guide](https://packaging.python.org/tutorials/distributing-packages/).

- use [setuptools_scm](https://github.com/pypa/setuptools_scm/) to generate version number, normal releases will use the tagged version (ie 1.9.1), versions installed from source will append a local version identifier[1] using the current git commit (ie 1.9.1.dev61+g4584c93). This means the version doesn't need manually updated, it will be generated when you build a package (ie `python setup.py sdist`) ~~or when you push a tag to travis-ci~~.

- use io.open() to read text files to correctly handle utf-8 characters

- use pip to parse requirements.txt

- update classifiers to specify python and django versions supported

- fix classifier for license

- allow universial wheel packages [2] to be built using `python setup.py bdist_wheel`

[1] [PEP 440](https://www.python.org/dev/peps/pep-0440/)
[1] https://packaging.python.org/tutorials/distributing-packages/#local-version-identifiers
[2] https://packaging.python.org/tutorials/distributing-packages/#wheels

@tonioo the license classifier in setup.py was set to `License :: OSI Approved :: MIT License` I assume that should have been `License :: OSI Approved :: ISC License (ISCL)` as per the LICENSE?

Any reason why travis doesn't build PyPi packages for modoboa when a tag is pushed like it does for the extensions?

If you want I'll add another commit with the travis settings?

```yaml
before_deploy:
  - cd ../modoboa
  - django-admin compilemessages
  - cd ..

deploy:
  provider: pypi
  user: XXX
  password:
    secure: XXX
  skip_cleanup: true
  distributions: "sdist bdist_wheel"
  on:
    tags: true
    python: "3.6"
    condition: $DB = POSTGRESQL
```


fyfe:
-0.03% 😠 for something that can only be tested if the package isn't installed, might have to adjust codecov so its failure thershold is a little higher, -0.05 maybe 😃

tonioo:
Didn't know pip was capable of parsing version numbers, that far better than the package I found. I didn't add the auto pypi release stuff to the modoboa repository because it is still a work in progress :p I had troubles to find the appropriate to store passwords but now I know! I'll add it.
You can update the LICENSE classifier, I simply forgot to do it...
Don't worry about codecov, I guess we can merge this PR even if this check fails?

setup.cfg
```diff
@@ -1,3 +1,5 @@
+[bdist_wheel]
+universal=1
```
tonioo: ```universal = 1``` to be consistent ?

fyfe: fixed

PR #1357, opened 2018-01-15
refactor cryptutils

fyfe:
Cryptography can be a black art, rather than rolling our own crypto
functions use the ones provided by the Python Cryptography module.

- `encrypt()` and `decrypt()`
    - ECB is insecure and the Python Cryptography module recommends[1]
    against using it.  The Python Cryptography module offers Fernet[2]
    for encrypting values using a secret_key.

    - Djangos `settings.SECRET_KEY` exists for things like encrypting
      passwords, use it instead of Modoboas own `secret_key`[3].

    - `settings.SECRET_KEY` is truncated to 32 characters as Fernet requires
       a key 32 characters long.

- `random_key()` uses Dangos get_random_string() method, the same
code used to generate settings.SECRET_KEY

- add tests for cryptutil methods

[1] https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/?highlight=ecb#insecure-modes
[2] https://cryptography.io/en/latest/fernet/
[3] It's my intention to remove Modoboas `secret_key` as it can be replaced by Djangos `settings.SECRET_KEY`, I just need to make sure it won't break modoboa-pdfcredentials as it also uses it.

### DIFF
--- a/modoboa/lib/cryptutils.py
+++ b/modoboa/lib/cryptutils.py
@@ -1,58 +1,49 @@
-# coding: utf-8
+# -*- coding: utf-8 -*-
+
 """Crypto related utilities."""
 
 from __future__ import unicode_literals
 
 import base64
-import random
-import string
 
-from django.utils.encoding import force_bytes, force_text
+from django.conf import settings
+from django.utils.crypto import get_random_string
+from django.utils.encoding import smart_bytes, smart_text
 
-from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
-from cryptography.hazmat.backends import default_backend
-
-from modoboa.parameters import tools as param_tools
+from cryptography.fernet import Fernet
 
 
-def random_key(l=16):
-    """Generate a random key.
-
-    :param integer l: the key's length
-    :return: a string
-    """
-    punctuation = """!#$%&'()*+,-./:;<=>?@[]^_`{|}~"""
-    population = string.digits + string.ascii_letters + punctuation
-    while True:
-        key = "".join(random.sample(population * l, l))
-        if len(key) == l:
-            return key
+def random_key(length=16):
+    """Generate a new key used to encrypt user passwords in session storage."""
+    chars = 'abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)'
+    return get_random_string(length, chars)
 
 
-def encrypt(clear):
-    key = force_bytes(
-        param_tools.get_global_parameter("secret_key", app="core"))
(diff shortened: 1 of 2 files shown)


modoboa/lib/cryptutils.py
```diff
@@ -1,58 +1,47 @@
...
+    fernet = Fernet(secret_key)
+    return smart_text(fernet.decrypt(smart_bytes(encrypted_value)))
 
 
 def get_password(request):
-    return decrypt(request.session["password"])
+    """
+    Retrieve and decrypt the users password from session storage.
+
+    This is used by modoboa-webmail to allow the user to send/recieve e-mails
```
tonioo: receive

modoboa/lib/tests/test_cryptutils.py
```diff
@@ -0,0 +1,40 @@
...
+class CryptUtilsTest(SimpleTestCase):
+
+    """Tests for modoboa.lib.cryptutils"""
+
+    def test_valid_secret_key_generated(self):
+        """Generate a 32 byte key"""
+        secret_key = cryptutils.random_key(32)
+        self.assertEqual(len(secret_key), 32)
+
+    @override_settings(SECRET_KEY="!8o(-dbbl3e+*bh7nx-^xysdt)1gso*%@4ze4-9_9o+"
```
tonioo: You could apply the override_settings decorator to the class instead of each method?

modoboa/lib/cryptutils.py
```diff
@@ -1,58 +1,47 @@
...
+    """Encrypt a value using secret_key."""
+    secret_key = base64.urlsafe_b64encode(smart_bytes(settings.SECRET_KEY[:32]))
+    fernet = Fernet(secret_key)
+    return smart_text(fernet.encrypt(smart_bytes(clear_value)))
+
+
+def decrypt(encrypted_value):
+    """Decrypt a value using secret_key."""
+    secret_key = base64.urlsafe_b64encode(smart_bytes(settings.SECRET_KEY[:32]))
+    fernet = Fernet(secret_key)
```
tonioo: Looks like you could create a ``get_fernet()`` function for that?

fyfe: fixed

modoboa/lib/cryptutils.py
fyfe: fixed

modoboa/lib/tests/test_cryptutils.py
fyfe: fixed

PR #1310, opened 2017-11-25
Include Fix from modoboa-webmail

bot315:
https://github.com/modoboa/modoboa-webmail/issues/121

PR #1360, opened 2018-01-18
improve e-mail address splitting functions

fyfe:
I've kept `split_address()` and `split_local_part()` for the modoboa-amavis code to work.

Needed for modoboa/modoboa-amavis#89

### DIFF
--- a/modoboa/lib/email_utils.py
+++ b/modoboa/lib/email_utils.py
@@ -10,6 +10,7 @@ import re
 import smtplib
 import time
 
+from django.utils import six
 from django.utils.encoding import smart_str, smart_text
 from django.utils.html import conditional_escape, escape
 from django.template.loader import render_to_string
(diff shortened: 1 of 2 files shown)


fyfe:
I've added a fix for #1310

modoboa/lib/email_utils.py
```diff
@@ -262,6 +263,47 @@ def render_headers(self, **kwargs):
...
+
+
+def split_local_part(local_part, delimiter=None):
+    """Split a local part into local part and extension."""
+    assert isinstance(local_part, six.text_type),\
+        "local_part should be of type %s" % six.text_type.__name__
+    assert isinstance(delimiter, six.text_type) or delimiter is None,\
+        "delimiter should be of type %s" % six.text_type.__name__
+    extension = None
+    if local_part.lower() in ["postmaster", "mailer-daemon", "double-bounce"]:
```
tonioo: Is the test against postmaster really necessary?

fyfe: fixed

modoboa/lib/email_utils.py
```diff
@@ -262,6 +263,47 @@ def render_headers(self, **kwargs):
...
+
+
+def split_local_part(local_part, delimiter=None):
+    """Split a local part into local part and extension."""
+    assert isinstance(local_part, six.text_type),\
+        "local_part should be of type %s" % six.text_type.__name__
+    assert isinstance(delimiter, six.text_type) or delimiter is None,\
+        "delimiter should be of type %s" % six.text_type.__name__
+    extension = None
+    if local_part.lower() in ["mailer-daemon", "double-bounce"]:
```
kryskool: These entries should be set as settings, because there may be custom addresses ?
eg:
LOCAL_PART_IGNORE = ["mailer-daemon", "double-bounce"]

fyfe: Good suggestion, however mailer-daemon, double-bounce, owner-* and *-request are the only special addresses that postfix and amavis deal with.

See http://www.postfix.org/postconf.5.html#recipient_delimiter

tonioo: I think we can start like this and make it configurable if we receive requests.

modoboa/lib/email_utils.py
```diff
@@ -262,6 +263,47 @@ def render_headers(self, **kwargs):
...
 
+def split_address(address):
+    """Split an e-mail address into local part and domain."""
+    assert isinstance(address, six.text_type),\
+        "address should be of type %s" % six.text_type.__name__
+    if "@" not in address:
+        local_part = address
+        domain = None
+    else:
+        local_part, domain = address.split("@", 1)
```
tonioo: I think we should consider the latest @, not the first. (example : "test@toto.com"@domain.tld) And actually, it was implemented like this before.

fyfe: fixed

PR #1366, opened 2018-01-26
Removed obsolete secret key.

tonioo:


fyfe:
@tonioo  this is still used by modoboa-pdfcredentials

https://github.com/modoboa/modoboa-pdfcredentials/blob/5827416c7a9da83d989fffb59872967772cbdef3/modoboa_pdfcredentials/lib.py#L50

tonioo:
Indeed, I'll made a PR for this plugin too.

fyfe:
@tonioo you should be able to replace it with `smart_bytes(settings.SECRET_KEY[:32])`

PR #1384, opened 2018-01-31
email utils error handling

fyfe:
with modoboa/modoboa-amavis#99 it fixes modoboa/modoboa-amavis#94

### DIFF
--- a/modoboa/lib/email_utils.py
+++ b/modoboa/lib/email_utils.py
@@ -10,15 +10,18 @@ import re
 import smtplib
 import time
 
+import chardet
+import lxml.html
+from lxml.html.clean import Cleaner
+
+from django.template.loader import render_to_string
 from django.utils import six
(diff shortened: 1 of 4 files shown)


modoboa/lib/email_utils.py
```diff
@@ -166,7 +169,24 @@ def _parse_text(self, msg, level=0):
...
+                    )
+                except (TypeError, UnicodeDecodeError) as exc:
+                    six.raise_from(
+                        InternalError(
+                            _(
+                                "unable to determine encoding of e-mail message"
+                            )
+                        ),
+                        exc
+                    )
```
tonioo: This code looks very similar to the one you put in the quarantine PR. There is also a ``decode()`` function is th same file. Maybe we could refactor them a bit?

fyfe: good point, refactored

Issue #1417, opened 2018-03-03
Psycopg2 wheel package will be renamed from release 2.8


fyfe:
From psycopg2 version 2.8 the binary wheel package will be renamed to `psycopg2-binary` (http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/).

This is due to some people having issues with Python and psycopg2 being compiled against different versions of libssl.

---

At the moment there haven't been any reported issues with Modoboa and the psycopg2 binary package so I'll create a PR to update `requirements.txt` to use the new binary package name.  Long term it might be best to use the source package I'll need to do some testing.

PR #1418, opened 2018-03-03
#1417 psycopg2 binary package has been renamed to psycopg2-binary

fyfe:
also bump the minimum version to 2.7.4 as this is when the psycopg2-binary became available.

See #1417 and http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/


tonioo:
@fyfe Thank you. We also need to update the installer.

PR #1426, opened 2018-03-06
[docs] remove uwsgi warning

cptMikky:
Current Debian stable ships with uWSGI 2.0.14 while the current upstream version to the date is only 2.0.17. The original warning no longer applies, there's no need to scare off Debian or Ubuntu users.


tonioo:
Indeed, thank you.

Issue #1430, opened 2018-03-08
CSV import broken


almereyda:
# Impacted versions

* Modoboa: 10.10.2
* installer used: Yes
* Webserver: Nginx

# Steps to reproduce

Run an email import from a CSV formed after http://modoboa.readthedocs.io/en/latest/moving.html#using-csv-files, i.e. one that worked before.

# Current behavior

- In the web frontend, the spinner will just rotate indifinitely.
- On the command line, an error is thrown:

```
(env)email# python /srv/modoboa/instance/manage.py modo import accounts.csv
/srv/modoboa/env/local/lib/python2.7/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)
Traceback (most recent call last):                                              |ETA:  --:--:--
  File "/srv/modoboa/instance/manage.py", line 22, in <module>
    execute_from_command_line(sys.argv)
  File "/srv/modoboa/env/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/srv/modoboa/env/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/srv/modoboa/env/local/lib/python2.7/site-packages/subcommand/base.py", line 53, in run_from_argv
    return super(SubcommandCommand, self).run_from_argv(argv)
  File "/srv/modoboa/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/srv/modoboa/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/srv/modoboa/env/local/lib/python2.7/site-packages/subcommand/base.py", line 84, in handle
    return command.run_from_argv(argv)
  File "/srv/modoboa/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/srv/modoboa/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
[...]

PR #1432, opened 2018-03-11
fix import command

fyfe:
I can't replicate the issue in #1430, but here's a couple of changes  that might fix it.

- ensure command line args are unicode(py2)/str(py3)

- use io.open to open file with utf8 encoding

- set newlines="" in open commands, csv module takes care of line endings

- add tests for import command

### DIFF
--- a/modoboa/admin/management/commands/subcommands/_import.py
+++ b/modoboa/admin/management/commands/subcommands/_import.py
@@ -8,9 +8,12 @@ import io
 import os
 
 import progressbar
+from chardet.universaldetector import UniversalDetector
 
 from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
 from django.utils import six
+from django.utils.translation import ugettext as _
 
 from modoboa.core import models as core_models
 from modoboa.core.extensions import exts_pool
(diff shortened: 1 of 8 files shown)


modoboa/admin/management/commands/subcommands/_import.py
```diff
@@ -41,19 +42,26 @@ def add_arguments(self, parser):
 
     def _import(self, filename, options):
         """Import domains or identities."""
+        if isinstance(filename, six.binary_type):
+            filename = filename.decode("utf-8")
+        if isinstance(options["sepchar"], six.binary_type):
+            options["sepchar"] = options["sepchar"].decode("utf-8")
```
tonioo: Why not use smart_str instead?

modoboa/admin/management/commands/subcommands/_import.py
```diff
@@ -41,19 +42,26 @@ def add_arguments(self, parser):
...
         superadmin = (
             core_models.User.objects.filter(is_superuser=True).first()
         )
-        if not os.path.exists(filename):
+        if not os.path.isfile(filename):
             raise CommandError("File not found")
 
-        num_lines = sum(1 for line in open(filename) if line)
+        num_lines = sum(
+            1 for line in io.open(filename, encoding="utf-8") if line
```
tonioo: Maybe we should try to detect file encoding before? (using chardet for example)

fyfe:
Well I've added chardet to the mix, the only issue is I've been unable to create a CSV file that can't be read as utf-8.

So I've not been able to write a test for it yet.

Issue #1438, opened 2018-03-19
pip 10 (due April) breakages


fyfe:
pip 10 is due in April and they've chosen to refactor all of pip's code to `pip._internal` to make it clear they don't want people using pip directly.

Modoboa depends on pip in 2 places:
https://github.com/modoboa/modoboa/blob/95ccae54ca72ac446c342257bdd22b312bb228f0/setup.py#L24
This can be fixed by using the [packaging](https://pypi.python.org/pypi/packaging) package.

https://github.com/modoboa/modoboa/blob/95ccae54ca72ac446c342257bdd22b312bb228f0/modoboa/core/commands/deploy.py#L231
This can be replaced by calling pip the same way we call any other external program.

PR #1439, opened 2018-03-19
remove dependency on pip internals

fyfe:
see #1438

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,13 +11,12 @@ coreapi==2.3.3
 coreapi-cli==1.0.6
 djangorestframework==3.7.3
 
-argparse
 # bcrypt requires libffi-dev
 bcrypt==3.1.4
 dnspython==1.15.0
 feedparser==5.2.1
 gevent==1.2.2
-ipaddress
+ipaddress; python_version < '3.3'
 jsonfield==2.0.2
 passlib==1.7.1
 progressbar33==2.4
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -11,13 +11,12 @@ coreapi==2.3.3
...
 djangorestframework==3.7.3
 
-argparse
 # bcrypt requires libffi-dev
 bcrypt==3.1.4
 dnspython==1.15.0
 feedparser==5.2.1
 gevent==1.2.2
-ipaddress
+ipaddress; python_version < '3.3'
```
tonioo: Does it mean this package is not compatible with python >= 3.3? If so, we need a replacement...

fyfe:
Its included in python 3.3+

On March 22, 2018 2:45:47 PM UTC, Antoine Nguyen <notifications@github.com> wrote:
>tonioo commented on this pull request.
>
>
>
>>  # bcrypt requires libffi-dev
> bcrypt==3.1.4
> dnspython==1.15.0
> feedparser==5.2.1
> gevent==1.2.2
>-ipaddress
>+ipaddress; python_version < '3.3'
>
>Does it mean this package is not compatible with python >= 3.3? If so,
>we need a replacement...
>
>-- 
>You are receiving this because you were assigned.
>Reply to this email directly or view it on GitHub:
>https://github.com/modoboa/modoboa/pull/1439#pullrequestreview-106149762

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #1447, opened 2018-03-31
ensure utf8 encoding is used when creating/accessing test database

fyfe:


modoboa/test_settings.py
```diff
@@ -26,7 +26,15 @@
             # See:
             # https://docs.djangoproject.com/en/1.11/ref/databases/#mysql-sql-mode
             "OPTIONS": {
-                "init_command": 'SET sql_mode=\'STRICT_TRANS_TABLES\'',
+                "init_command": (
+                    "SET sql_mode = STRICT_TRANS_TABLES;"
```
tonioo: Why have you removed the quotes?

fyfe: It works with out quotes, but just to be safe and as the MySQL documentation examples use quotes I've added them back in. However `innodb_strict_mode` should not have quotes as `ON` is a boolean value in MySQL.

PR #1452, opened 2018-04-06
[WIP] refactor setyp.py and requirements.txt

fyfe:
This is a [better solution](https://packaging.python.org/discussions/install-requires-vs-requirements/) for #1438 and also fixes modoboa/modoboa-installer#197.  

This isn't ready for merging yet it still needs testing/reviewing.


setup.py
```diff
@@ -76,12 +50,54 @@ def get_requirements(requirements_file):
         include_package_data=True,
         zip_safe=False,
         scripts=["bin/modoboa-admin.py"],
-        install_requires=INSTALL_REQUIRES,
+        install_requires=[
```
kryskool: Why not move these to the [setup.cfg](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files) file ?

setup.py
```diff
@@ -76,12 +50,54 @@ def get_requirements(requirements_file):
...
+            "backports.csv; python_version < '3.3'",
+            "chardet",
+        ],
         use_scm_version=True,
         setup_requires=["setuptools_scm"],
         extras_require={
-            "ldap": LDAP_REQUIRES,
-            "mysql": MYSQL_REQUIRES,
-            "postgresql": POSTGRESQL_REQUIRES,
+            "ldap": [
```
kryskool: Move to setup.cfg (see above)

fyfe:
stick with what we've got for now, this solution just creates more duplication.

PR #1464, opened 2018-04-12
PEP508 markers in requirements.txt have been broken since dfabe89

fyfe:
fixes modoboa/modoboa-installer#197


kryskool:
@fyfe thanks for the report and fix

Regards,

PR #1466, opened 2018-04-12
[Django 2.0]  Django instance namespace without app namespace is depreciated

fyfe:
Fix the last `RemovedInDjango20Warning`.

`url_namespace_required ` in `ModoExtension` provides compatibility until all the extensions have been fixed then the `url_namespace_required ` code can be removed.

See https://docs.djangoproject.com/en/1.11/ref/urls/#include

### DIFF
--- a/modoboa/admin/urls.py
+++ b/modoboa/admin/urls.py
@@ -12,6 +12,8 @@ from .views import (
     identity as identity_views, import_ as import_views
 )
 
+app_name = "admin"
+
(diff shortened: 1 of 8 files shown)


tonioo:
@fyfe Looks like we need to shutdown some codacy warnings...

fyfe:
sorry for the noise, codacy should behave itself now and not spam PRs with comments.

PR #1551, opened 2018-08-08
Docs for manual installation: Added update_statistics call to cron.

jakicoll:
Docs for manual installation are missing the update_statistics call in crontab.

It has been added to the installer with [this commit](https://github.com/modoboa/modoboa-installer/commit/6c6945c11b8763851732041e3c2fbd4abda44107) so I guess it should be added here too.


tonioo:
Thank you.

PR #1558, opened 2018-08-29
Cant Upload Attachment ( FIle too Big)

makakken:
size2 integer was returning 0 when called with int values.

https://github.com/modoboa/modoboa-webmail/issues/170

Current behavior before PR:
size2 integer was returning 0 when called with int values.

size2 integer returns the int value it was called with.


makakken:
Well, maybe you could tell me where i can put a unit-test-case to test the behavior?
It's my first contribution in python, i do not know if this is very elegant, so let me know if i could do something better.

PR #1578, opened 2018-10-05
feat(useDnsServer): add field in admin view, use in resolver if defin…

fvandenbulcke:
Add in admin, field to use specific DNS server.
The server is call during query resolving.
Unit tests for view and method about query resolving

### DIFF
--- a/modoboa/admin/app_settings.py
+++ b/modoboa/admin/app_settings.py
@@ -83,7 +83,7 @@ class AdminParametersForm(param_forms.AdminParametersForm):
         label=ugettext_lazy("Use specific DNS server"),
         required=False,
         help_text=ugettext_lazy(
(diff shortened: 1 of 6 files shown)


modoboa/admin/app_settings.py
```diff
@@ -78,6 +79,14 @@ class AdminParametersForm(param_forms.AdminParametersForm):
         )
     )
 
+    use_specific_dns_server = GenericIPAddressField(
+        label=ugettext_lazy("Use specific DNS server"),
+        required=False,
+        help_text=ugettext_lazy(
+            "Give specific DNS server instead of local server configuration"
```
tonioo: Use a specific DNS server instead of local server configuration

tonioo:
@fvandenbulcke Thank you very much for your time ! :-)

Issue #1584, opened 2018-10-12
unable to login when I change the default password scheme


jodumont:
Hi;

1st: thanks for your marvelous project; it's fit so much my expectation _(no docker, postgres, ...)_
2nd: it's more a curiously thing than an angry issue but...

I change the default password scheme right after the installation, before creating any domain, user, ...
then I made the domain and user

the user is/was able to connect to the modoboa interface but
the webmail was "broken" and it was impossible to connect via IMAP & SMTP
so to fix it I put back the SHA512 and reset the password than everything works fine.
so I presume the password scheme don't interact with dovecot and postfix.

# Impacted versions

* Modoboa: 1.10.7
* installer used: Yes
* Webserver: Nginx
* Fresh Debian9 from a netinstall with only ssh (nothing else (no basic tools, ...))

# Steps to reproduce
1. goto : https://mail.yourdomain.tld/core/#parameters/
2. change the **Default password scheme** for `bcrypt`
3. define a user
4. login as this user
5. try webmail

# Current behavior
I survive by putting back `SHA512`

# Expected behavior
being able to use **bcrypt** because it is offered and also because [this article](https://rietta.com/blog/2016/02/05/bcrypt-not-sha-for-passwords/)

tonioo:
Looks like dovecot (< 2.3) relies on local libc to provide BCRYPT (see https://wiki.dovecot.org/Authentication/PasswordSchemes). Unfortunately, BCRYPT is not available on Debian 9 and it provides dovecot 2.2... (see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=149452).
I think the only fix we can propose is to hide schemes that are not available on the local server. Then, we will propose new schemes.

jodumont:
@Toniob yes yes I put it back to SHA512-CRYPT, the one by default
not SHA512 which -CRYPT was lost in the translation

@tonioo don't worry about it; as I said; I was more curious about it than anything else.

PR #1587, opened 2018-10-13
check for dovecot supported password schemes

Toniob:
call doveadm pw -l to list password schemes
only show supported password schemes


modoboa/core/password_hashers/__init__.py
```diff
@@ -29,3 +34,26 @@ def get_password_hasher(scheme):
...
+    supported_schemes = ["{PLAIN}"]
+    dpath = None
+    code, output = exec_cmd("which doveadm")
+    if not code:
+        dpath = force_text(output).strip()
+    else:
+        known_paths = ("/usr/bin/doveadm", "/usr/local/bin/doveadm")
+        for fpath in known_paths:
+            if os.path.isfile(fpath) and os.access(fpath, os.X_OK):
+                dpath = fpath
```
tonioo: You can add a break once you've found a match.

modoboa/core/password_hashers/__init__.py
```diff
@@ -29,3 +34,26 @@ def get_password_hasher(scheme):
...
+    if dpath:
+        try:
+            code, schemes = exec_cmd("%s pw -l" % dpath)
```
tonioo: no .format() here?

modoboa/core/password_hashers/__init__.py
```diff
@@ -29,3 +34,26 @@ def get_password_hasher(scheme):
...
+
+def get_dovecot_schemes():
+    """Return schemes supported by dovecot"""
+    supported_schemes = ["{PLAIN}"]
+    dpath = None
+    code, output = exec_cmd("which doveadm")
+    if not code:
+        dpath = force_text(output).strip()
+    else:
+        known_paths = ("/usr/bin/doveadm", "/usr/local/bin/doveadm")
```
tonioo: I think there is another place in the code where we look for doveadm path. I guess a new function would be welcome? (maybe in lib/sysutils.py)

modoboa/core/password_hashers/advanced.py
```diff
@@ -49,6 +52,9 @@ class MD5CRYPTHasher(PasswordHasher):
 
     This scheme can't be considered as secure anymore.
     """
+    name = "md5crypt"
+    label = "md5crypt"
```
tonioo: Two properties in the base class would be better. The scheme name can be guessed from the class name and we could add a weak flag too.

Toniob:
closes #1584

PR #1608, opened 2018-11-01
Pass parameters for shell function call using an array rather than (unquoted) string substitution

alexander255:
Modoboa assembles strings for shell commands using string substitution without quoting the parameters, leading, at least potentially (I haven't checked), to vulnerabilities (shell injection).

Current behavior before PR:
Assembling shell parameter lists was broken.

Assembling shell parameter lists should not be vulnerable to shell injection attacks.


alexander255:
I fixed the dumb codestyle checker and the bug reported by the CI. Codecov is wrong on its claim however, as there is no coverage change because of this patch: https://codecov.io/gh/modoboa/modoboa/pull/1608/diff

modoboa/lib/sysutils.py
```diff
@@ -22,16 +22,16 @@ def exec_cmd(cmd, sudo_user=None, pinput=None, capture_output=True, **kwargs):
...
     :param str pinput: data to send to process's stdin
     :param bool capture_output: capture process output or not
     :rtype: tuple
     :return: return code, command output
     """
     if sudo_user is not None:
-        cmd = "sudo -u %s %s" % (sudo_user, cmd)
-    kwargs["shell"] = True
+        cmd = ["sudo", "-u", sudo_user] + list(cmd)
+    kwargs["shell"] = False
```
tonioo: I don't think it's a good idea to change this value since you did not modify commands to include absolute pathes. If you disable shell mode, PATH resolution won't be available, right?

PR #1648, opened 2019-01-07
Bugfix: DKIM checks fail

jammo-nl:
The DKIM key is base64 encoded, meaning it can end with "=" signs
Split on "=" and check if you have 2 'parts' (key and value) will fail in these cases.
By using split("=",1) we make sure we only slit on the first occurence of the "=" sign
(solves modoboa#1647 and partially modoboa#1640 and modoboa#1642)

Current behavior before PR:
DKIM checks fail

DKIM checks work


tonioo:
@jammo-nl Thank you.

PR #1686, opened 2019-02-28
fix regression that caused the rounds parameter to be ignored

Arvedui:
A regression introduces about year ago caused the rounds parameter to be ignored

Current behavior before PR:
The rounds parameter is ignore, passlib always uses the default which for sha512_crypt is 656000.

The rounds parameter is honored.
You can verify this by setting it to != 656000 and the password hash to sha{256,512}_crypt.
Then update the pw and verify the result in the `password` field of the `core_user` table


tonioo:
@Arvedui Thank you!

PR #1701, opened 2019-03-19
Make dependency on argon2 optional

jeromelebleu:
## Description of the issue this PR addresses
Thanks to #1690, Argon2 is now supported as password hasher. However, it is not optional and Argon2 requires a third-party library - that's why Django do not use it as default (see the [related documentation](https://docs.djangoproject.com/en/2.1/topics/auth/passwords/#using-argon2-with-django)). This PR try to make this password hasher optional.

## Current behavior before PR
* argon2 is in the requirements
* argon2 is imported in modoboa/core/password_hashers/advanced.py
* ARGON2IDHasher is defined and inherits from PasswordHasher
* tests depend on argon2

* argon2 is not in requirements but in extras
* argon2 is imported in modoboa/core/password_hashers/advanced.py with a `try...except` catch
* ARGON2IDHasher is defined but only inherits from PasswordHasher if argon2 is installed
* tests related to argon2 are skipped if it is not installed

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,3 +1,4 @@
+argon2-cffi>=16.1.0
 factory-boy>=2.4
 mock==2.0.0; python_version < '3.3'
 httmock==1.2.5
--- a/modoboa/core/password_hashers/advanced.py
+++ b/modoboa/core/password_hashers/advanced.py
@@ -9,7 +9,11 @@ This module relies on `passlib` to provide more secure hashers.
 from __future__ import unicode_literals
 
 from passlib.hash import bcrypt, md5_crypt, sha256_crypt, sha512_crypt
-from argon2 import PasswordHasher as argon2_hasher
+
+try:
+    from argon2 import PasswordHasher as argon2_hasher
+except ImportError:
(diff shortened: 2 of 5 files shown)


test-requirements.txt
```diff
@@ -1,3 +1,4 @@
+argon2-cffi>=16.1.0; python_version >= '3'
```
tonioo: @jeromelebleu I don't remember why this dependency is only installed with python3?

jeromelebleu: I don't remember too as of now... As I didn't find any reason after a quick search - it is said that [argon2-cffi](https://pypi.org/project/argon2-cffi/) is tested with Python 2 - I suggest to remove the requirement on the Python version.

tonioo: Fine with me. Can you do it please?

jeromelebleu: It's done!

camdenorrb:
How do you enable argon2 after installation?

Arvedui:
The argon2 support has not yet been released, you would need to install current master and make sure the dependencies are installed. Then you can choose it in the settings.

Issue #1721, opened 2019-04-21
Remove bad.psky.me from DNSBL providers


yannfill:
You should remove bad.psky.me from DNSBL providers list. It was [fishy at first](https://www.spamhaus.org/organization/statement/015/fraudulent-dnsbl-uncovered-protected-sky-bad.psky.me) and now the domain seems to have been drop catched, with website displaying a parking page. It then results in false alert in Modoboa RBL check.

Lucidi0us:
Today I had about 13 emails from Modoboa about bad.psky.me DNSBL check.  I agree with @yannfill that this one needs to go.  It was considered fraudulent for a while (see: https://www.spamhaus.org/organization/statement/015/fraudulent-dnsbl-uncovered-protected-sky-bad.psky.me ).

ewenjo:
I agree with this. Got tons of emails yesterday that my private IPs were listed by them, which are false positives.

dugite-code:
For thoes waiting you can manually disable a DNSBL in the `settings.py by adding the following snipit:

```python
# fix broken DNSBLs
from modoboa.admin.constants import DNSBL_PROVIDERS
DNSBL_PROVIDERS.remove("bad.psky.me")
```
From: https://github.com/modoboa/modoboa/issues/1496#issue-326768587

dym-sh:
/rel https://hetrixtools.com/forum#/discussion/91/bad-psky-me-inactive-removed

almereyda:
Is the solution via `settings.py` to be preferred over the solution presented in the PRs #1722 #1723 #1724?

dugite-code:
@almereyda the `settings.py` solution is just while you wait to update to the next version of modoboa. It's just how you can customize you DSNBL for you individual system.

almereyda:
#1722 has been merged. Can this be closed?

warsix4:
Supposedly this is fixed, however I am getting notified about this every few hours still with a fresh installation. Any reason it's persisted still?

tonioo:
@warsix4 Because it has not been released yet.

PR #1722, opened 2019-04-22
Removed bad DNSBL bad.psky.me - see issue #1721.

Eveley:
Description of the issue/feature this PR addresses: Messages from the system about a domain being blacklisted at bad.psky.me when that is not correct, the provider is hijacked or is otherwise nonfunctional as of April 21.

Current behavior before PR:
Several messages a day to postmaster from Modoboa saying that a domain has been blacklisted at bad.psky.me.

No more messages from Modoboa about this blacklist provider as it is removed from checks.


tonioo:
@Eveley Thank you

PR #1724, opened 2019-04-24
remove bad.psky.me

y0va:
As it is not trustworthy. See:  
https://www.spamhaus.org/organization/statement/015/fraudulent-dnsbl-uncovered-protected-sky-bad.psky.me

For us the problem was that we had a lot of false positives of our own domains with it.


almereyda:
It seems this is a duplicate of #1722.

tonioo:
See #1722. Thank you.

Issue #1496, opened 2018-05-26
Private DNSBL listed in DNSBL check: dnsbl.cyberlogic.net


whyscream:
# Impacted versions

* Modoboa: git HEAD
* Installer used:  irrelevant
* Webserver: irrelevant

# Steps to reproduce
Look at email box, see that I have received various emails about a DNSBL that is broken, badly maintained , irrelevant, and I never knew existed.

# Current behavior
The list of DNSBLs defined at `modoboa/admin/constants.py` includes `"dnsbl.cyberlogic.net"`, which is currently broken (see: https://www.dnsbl.com/2018/05/status-of-dnsblcyberlogicnet-broken.html). I never heard of this DNSBL, and the author of referenced article hasn't either. Based on his findings I'm assuming that this DNSBL was actually private.

# Expected behavior
1) I would not expect Modoboa to include such obscure DNSBLs by default, just a list of widely accepted and properly supported lists. Please remove the mentioned DNSBL from the modoboa defaults, and any other non-public DNSBLs too. For instance: cyberlogic doesn't have a website listed at https://www.dnsbl.info/dnsbl-details.php?dnsbl=dnsbl.cyberlogic.net, so performing manual lookups, reading about listing policy and/or getting a removal is a pain to say the least.
2) Proper documentation on how to remove DNSBLs from the list. I now found out (by reading the code) that Modoboa actually has quite nice support for removing/customizing this list. I fixed my issue by adding the following snippet in my settings, but this should be available for anyone reading documentation.

```python
# fix broken DNSBLs
from modoboa.admin.constants import DNSBL_PROVIDERS
DNSBL_PROVIDERS.remove("dnsbl.cyberlogic.net")
````

oconshaw:
Great find and thanks for posting a fix.  I am having the same issue. Can you elaborate a bit more on how to apply the fix you detailed above?  Where/how do you add that snippet to your "settings". --Thanks.

whyscream:
After installing, the Modoboa installation created a few files. One of them is named `settings.py`. you can append the lines to that file.

almereyda:
In our instance, provisioned by the `modoboa-installer`, the file is living in `/srv/modoboa/instance/instance/settings.py`.

A `service uwsgi restart` usually helps picking up those changes.

TacoScheltema:
following up on this; bl.spamcannibal.org recently had it's domain name hijacked and is now serving up advertisements and malicious content. This blacklist should also be removed from modoboa.

[...]

PR #1748, opened 2019-06-14
Remove lots of dead and/or unreliable DNSBLs from the default list.

whyscream:
Clean the default list of DNSBLs used by the DNSBL checks. The existing list os old and unmaintained, and lists many dead or unmaintained DNSBLs.

I shrunk the list to a short list of well-known, widely used DNSBLs that respect proper listing policies, which means that when Modoboa signals a user that an ipaddress is listed, the user should be able to find out what went wrong, and how to resolve that. 

This ensures that users don't get false positives about DNSBL checks because unmaintained lists go dark or get hijacked, and they can act when an actual listing of their server occurs.

Most of this was discussed earlier in #1496.


tonioo:
@whyscream Thank you!

PR #1790, opened 2019-11-12
Fix exception handling in argon2 for mismatching passwords

Arvedui:
Fix exception handling in argon2 for mismatching passwords

Current behavior before PR:
When the argon2 hasher is used and the given password does not match the hash, an "internal error" occurs.

The proper "invalid username or password" message is shown and now redirect to the error page happens.


tonioo:
@Arvedui Thank you for the fix but could you please retrieve the last fixes from master? I've removed Python3.4 from tests matrix.

Arvedui:
done

PR #1796, opened 2019-12-01
FIX: relaydomain: next hop should matched with the domain's actual MX list

tsabi:
FIX: relaydomain: next hop should matched with the domain's actual MX list

If you create a relaydomain, than the DKIM and SPF checks are useful,
because they check that in the domain's data you really have good
config. But the MX check is not useful, because it tries to match with
your valid MX list (setted globally in the config panel), what is a good
behavior for a handled domain, but for a relay domain it is not.

This modification modifies the MX record checking methodology that it
checks if your relay hop is setted as a MX receiver for that domain, so
with the MX red/green tag on your dashboard you can check that you
forward your emails into the good target server or not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ modoboa_test
 .coverage
 build/
 dist/
+/.project
+/.pydevproject
--- a/modoboa/admin/management/commands/subcommands/_mx.py
+++ b/modoboa/admin/management/commands/subcommands/_mx.py
@@ -155,6 +155,9 @@ class CheckMXRecords(BaseCommand):
         if not mxs:
             alerts.append(_("Domain {} has no MX record").format(domain))
         elif valid_mxs:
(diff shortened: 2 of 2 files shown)


tonioo:
@tsabi Do you know the valid MX list setting is not mandatory?

tsabi:
@tonioo Yes i know. I think if the MX checking is disabled at all, than it is not a problem if the relaydomain's MX validity will not be checked, because than the normal domain's MX validity will not be checked too.
Or i understanded you and is it something else you would like to ask from me to fix in my code?

tonioo:
@tsabi the setting you're talking about is just here to restrict the list of authorized MX IPs. If one of the domain hosted in modoboa uses an IP not listed here, it will trigger an alert. If the list is empty, modoboa will just check if every domain has a valid MX record declared and will not check the IP. Is it how you understood this setting?

PR #1811, opened 2019-12-29
Update tox.ini and tests to be run locally

jeromelebleu:
This fixes some cases where we want that tox passes all tests locally and updates Python versions to check.

### DIFF
--- a/modoboa/ldapsync/tests.py
+++ b/modoboa/ldapsync/tests.py
@@ -4,8 +4,6 @@ from __future__ import unicode_literals
 
 from unittest import skipIf
 
-import ldap
-
 from django.utils import six
 from django.utils.encoding import force_bytes, force_str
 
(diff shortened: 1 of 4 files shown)


modoboa/ldapsync/tests.py
```diff
@@ -4,16 +4,20 @@
...
 from unittest import skipIf
 
-import ldap
-
 from django.utils import six
 from django.utils.encoding import force_bytes, force_str
 
 from modoboa.core import factories as core_factories
-from modoboa.lib.tests import NO_LDAP, ModoTestCase
+from modoboa.lib.tests import ModoTestCase
```
tonioo: What was the issue with this import?

jeromelebleu: No one, it is just set instead just below. Do you prefer to rely on `modoboa.lib.tests.NO_LDAP` to just import or not `ldap`  and `.lib` - e.g. `if not NO_LDAP: import...`?

tonioo: Yes, I think it's better to keep the old way.

PR #1812, opened 2020-01-02
Remove Python 2 support

jeromelebleu:
Python 2 support has ended yesterday - i.e. 2020-01-01... This drops Python 2.7 from Modoboa and update the code to move to Python 3 only:
* Update each dependency to the most recent version which doesn't require big changes - i.e. *Django* and *Django REST Framework* are not updated to the last release because of that
* Remove coding header and `__future__` imports
* Remove use of `six`
* In `setup.py`, remove use of `io.open`, add `python_requires` and update classifiers
* In README.rst, specify the Python 3 version support
* In `.travis.yml` and `tox.ini`, stop testing on Python 2
* In `setup.cfg`, remove `[bdist_wheel]` section


jeromelebleu:
I had a look at the coverage decrease and I think it is probably a side effect of the `from __future__ import ...` and other lines removal.

tonioo:
@jeromelebleu is this PR ready to merge?

jeromelebleu:
Yes! #1818 may be merged first, but it doesn't really matter.

tonioo:
@jeromelebleu Great. Thank you!

PR #1822, opened 2020-01-13
Bump Django to 2.2

jeromelebleu:
Please see the commit messages for details about each update:
* [x] Bump djangorestframework to 3.11 and move to OpenAPI (e0c7305)
* [x] Bump Django to 2.0 and simplify URL routing syntax (8170156) - and fix #1466
* [x] Bump Django to 2.1 and rename *view* permissions (8d1f1e5)
* [x] Bump Django to 2.2 (0968a1c)

The changes that must be considered and which could affect extensions are:
* In Django REST Framework, OpenAPI is now used instead of CoreAPI to generate the schema
* Only `view_domain` is kept to check permissions - instead of both `view_domain` and `view_domains`
* `view_aliases`, `view_domaliases` and `view_mailboxes` are renamed to singular form - i.e. `view_alias` -` to match the [new *view* permission](https://docs.djangoproject.com/en/stable/releases/2.1/#model-view-permission)


jeromelebleu:
@tonioo It is now ready for your precious review!

modoboa/admin/migrations/0001_initial.py
```diff
@@ -19,7 +19,6 @@ class Migration(migrations.Migration):
             ],
             options={
                 'ordering': ['address'],
-                'permissions': (('view_aliases', 'View aliases'),),
```
tonioo: Old migrations must remain untouched, otherwise fresh installations won't behave like new ones and it could lead to unexpected migration issues.

modoboa/admin/migrations/0015_rename_view_permissions.py
```diff
@@ -0,0 +1,44 @@
...
+def rename_view_permissions(apps, schema_editor):
+    Permission = apps.get_model('auth', 'Permission')
+    db_alias = schema_editor.connection.alias
+    # Iterate over permissions and try to rename the codename
+    for perm in Permission.objects.using(db_alias).filter(
+        codename__in=VIEW_PERMISSION_CODENAMES.keys()
+    ):
+        codename = perm.codename
+        try:
+            with transaction.atomic():
```
tonioo: Is it really necessary? I thought migrations were already run within a transaction context...

modoboa/bower_components/swagger-ui/.bower.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "swagger-ui",
```
tonioo: Why this new dependency?

modoboa/core/extensions.py
```diff
@@ -103,17 +103,15 @@ def get_urls(self, category="app"):
         result = []
         for ext_name in list(self.extensions.keys()):
             ext = self.get_extension(ext_name)
-            options = {}
             if category == "api":
                 root = ""
                 pattern = "{}.urls_api"
             else:
                 root = r"^{}/".format(ext.get_url())
-                options.update({"namespace": ext_name})
```
tonioo: Namespace is automatically added by Django now?

modoboa/limits/migrations/0001_squashed_0004_auto_20160413_1312.py
```diff
@@ -0,0 +1,56 @@
+# Generated by Django 2.0.13 on 2020-01-04 17:21
```
tonioo: Why did you add this migration?

modoboa/admin/migrations/0001_initial.py
jeromelebleu: I don't think there is an impact with this migration change. I have removed them for at least two reasons:
* as `view_aliases` - and others - is finally renamed to `view_alias` in the last migration, and because permissions are created at the end of the migration process, it prevents to create the `view_aliases` permission without renaming it on fresh installations.
* for `view_domain`, it prevents a conflict starting with Django 2.1 where this permission will be automatically created by Django. If you try to keep this one in the migration and run `./manage.py migrate`, you will end with a `UNIQUE constraint failed: auth_permission.content_type_id, auth_permission.codename`.

modoboa/admin/migrations/0015_rename_view_permissions.py
jeromelebleu: Yes, it is as soon as an `IntegrityError` is encountered to allow the other operations to be committed. Otherwise, a `TransactionManagementError` will be raised - see this [doc](https://docs.djangoproject.com/en/stable/topics/db/transactions/#django.db.transaction.atomic). I have also discovered that when I tried to see if it was working as excepted.

modoboa/bower_components/swagger-ui/.bower.json
jeromelebleu: It is needed and used to render the API documentation in `modoboa/templates/swagger-ui.html`. It comes with e0c7305149097cb40ef4b32641f93881e12071f3 and is explained in the [Django REST Framework doc](https://www.django-rest-framework.org/topics/documenting-your-api/). The other choice would be [ReDoc](https://github.com/Rebilly/ReDoc), but as Swagger UI is more generally used... Also, this is needed because of the drop of `CoreAPI` - which is needed to use [`include_docs_urls`](https://www.django-rest-framework.org/coreapi/from-documenting-your-api/#include_docs_urls).

modoboa/core/extensions.py
jeromelebleu: The namespace mechanism has changed and it now requires that the `app_name` is set in each extension instead - as it is also done in `modoboa.admin` and so. This was reported at first by #1466.

modoboa/limits/migrations/0001_squashed_0004_auto_20160413_1312.py
jeromelebleu: This new migration replaces `0001_initial` to `0004_auto_20160413_1312` and was needed to prevent some unexpected issues - and headache - starting with Django 2.0. I think that this is mainly because of the `AlterModelTable` in `0002_auto_20151114_1518`. As this model changes occurred in 2016, I proceeded from the assumption that there is no more instances without them applied. Anyway, we could add a note if it is not the case to tell that such old-old-old instances should be upgraded to v1.14.0 at first.

tonioo: Ok I see. And adding a note is a good idea.

modoboa/core/extensions.py
tonioo: Ok.

modoboa/bower_components/swagger-ui/.bower.json
tonioo: Ok, makes sense.

modoboa/admin/migrations/0015_rename_view_permissions.py
tonioo: Ok, good to know.

modoboa/admin/migrations/0001_initial.py
tonioo: Ok.

tonioo:
@jeromelebleu Is this PR ready to merge?

jeromelebleu:
Yes sure, if it is okay for you. We will just have to add all of that to the upgrade note, but I can handle that in a separate PR - which could sum up all the changes of the next release?

tonioo:
Sure, let's to this.

PR #1766, opened 2019-08-04
dnstools dkim TXT record evaluation fix

tsabi:
Description of the issue/feature this PR addresses:
The DKIM button was red, but my TXT record setted well. So i looked about the issue, and i founded the evaluation of the TXT record is not good. According to DNS standard, a TXT record what is longer than 256 character need to split to several strings: https://help.directadmin.com/item.php?id=552

Current behavior before PR:
I just merge the splitted parts together.

Desired behavior after PR is merged:
The DKIM button is now green as it should be.

PR #1823, opened 2020-01-13
Handle multiple strings in DNS record value

jeromelebleu:
Fix #1766 for all the RR types and with tests.


tonioo:
@jeromelebleu Great, thank you!

PR #1859, opened 2020-03-26
Adding setting to disable RSS requests.

### LINKED ISSUE #1797: Dashboard - server behind proxy
# Impacted versions

* Modoboa: 1.14.0
* Webserver: Nginx

# Steps to reproduce
Modoboa server is behind proxy, so no internet direct access
Acces dashboard via admin account

# Current behavior
504 Gateway Time-out

# Expected behavior
no error

ClemRiviere:
Resolves #1797 

Current behavior before PR:
Modoboa server is behind proxy, so no internet direct access
Acces dashboard via admin account

-> 504 Gateway Time-out

It is possible to disable RSS requests in settings.py file being behind a proxy to avoid the dashboard time-out.

### DIFF
--- a/modoboa/core/commands/templates/settings.py.tpl
+++ b/modoboa/core/commands/templates/settings.py.tpl
@@ -297,6 +297,8 @@ SILENCED_SYSTEM_CHECKS = [
     "security.W019",  # modoboa uses iframes to display e-mails
 ]
 
+DISABLE_RSS = False
+
 # Load settings from extensions
 {% for extension in extra_settings %}
 try:
--- a/modoboa/core/commands/templates/settings.py.tpl
+++ b/modoboa/core/commands/templates/settings.py.tpl
@@ -297,7 +297,7 @@ SILENCED_SYSTEM_CHECKS = [
     "security.W019",  # modoboa uses iframes to display e-mails
 ]
 
-DISABLE_RSS = False
+DISABLE_DASHBOARD_EXTERNAL_QUERIES = False
 
 # Load settings from extensions
 {% for extension in extra_settings %}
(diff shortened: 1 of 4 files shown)


tonioo:
@ClemRiviere I think an online setting would be better here. Have you seen how it works?

ClemRiviere:
@tonioo By "online setting" do you mean "in-app setting" ? Giving the fact we can't access the dashboard, I'm not sure it's a good solution to provide this setting directly into the in-app settings panel.

tonioo:
Yes I meant inline setting. I think you're right, let's stay like this.

tonioo:
@ClemRiviere Can you add unit tests please?

modoboa/core/views/dashboard.py
```diff
@@ -43,11 +45,13 @@ def get_context_data(self, **kwargs):
...
+        if not settings.DISABLE_RSS:
+            posts = feedparser.parse(feed_url)
         entries = []
-        for entry in posts["entries"][:5]:
-            entry["published"] = parser.parse(entry["published"])
-            entries.append(entry)
+        if not settings.DISABLE_RSS:
+            for entry in posts["entries"][:5]:
+                entry["published"] = parser.parse(entry["published"])
+                entries.append(entry)
```
tonioo, line 54: To avoid 2 checks, you could rewrite your code as follows:
```python
entries = []
if not settings.DISABLE_RSS:
    posts = feedparser.parse(feed_url)
    for entry in posts["entries"][:5]:
        entry["published"] = parser.parse(entry["published"])
        entries.append(entry)

File: modoboa/core/commands/templates/settings.py.tpl [hunk 2]
```diff
@@ -297,6 +297,8 @@ SILENCED_SYSTEM_CHECKS = [
     "security.W019",  # modoboa uses iframes to display e-mails
 ]
 
+DISABLE_RSS = False
```
tonioo, line 300: I think this name is the best one to describe what you're implementing in this PR. Something like `DISABLE_DASHBOARD_EXTERNAL_QUERIES` would be better in my opinion.

PR #1884, opened 2020-05-07
Convert Dockerfile to Multi-Stage

jjasonkal:
- Decrease Dockefile image from 637MB to 294MB, leveraging Docker Multi-Stage technique.


tonioo:
@jjasonkal Thank you!

PR #2039, opened 2020-10-21
Polish and Fixing issues when using "docker-compose up"

BigMichi1:
when these changes have been applied the command `docker-compose up` will not fail during the build anymore and also the login is possible afterwards (usable UI)

Current behaviour before PR:
- docker build fails because of missing dependencies
- login is not possible after docker container is running

Desired behaviour after PR is merged:
`docker-compose up` will provide a container which can be used


tonioo:
@BigMichi1 Thank you!

Issue #2064, opened 2020-10-24
Python 3.5 End Of Life


kryskool:
Some packages update not available in python 3.5, like

- feeparser
- dnspython
- jsonfield

python 3.5 has reached its end of life

Are there any objections to removing it from the automated tests?

kryskool:
Some depracated warning present:

```
CryptographyDeprecationWarning: Python 3.5 support will be dropped in the next release ofcryptography. Please upgrade your Python.

  from cryptography.fernet import Fernet
```

PR #2065, opened 2020-10-24
Deprecate Python 3.5 and Add Python 3.9

kryskool:
See #2064 

Remove Python 3.5 
Add Python 3.9


kryskool:
For Python 3.9 need to merge #2068  before (remove Python 2 compatibility)

kryskool:
Now we use Github Actions

PR #2094, opened 2020-12-04
Remove Reference to System Package

vthg2themax:
This will help people not be confused by the fact python3 is the new default.

Current behavior before PR:

### DIFF
--- a/doc/manual_installation/modoboa.rst
+++ b/doc/manual_installation/modoboa.rst
@@ -96,12 +96,6 @@ Install the corresponding Python binding:
 
    (env)$ pip install psycopg2
 
-.. note::
-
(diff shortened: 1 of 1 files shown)


MrGeneration:
Just a thought... is it safe to expect that the user that's trying to install modoboa is using the most current possible operating system that uses said standards?

On other projects I've seen people using far older versions (e.g. Ubuntu 16 over 20) just "because they had that already".

kryskool:
Hi @vthg2themax 

I need to reject this pull request, because it's better to use system package when you can, this part is just a note.

@MrGeneration if users use an old system, it's their problem and responsabilty. in the documetation we try to keep it up to date with the last distributions availables

Regards,

PR #2144, opened 2021-01-27
Prevent that password are stored in plain-text

jeromelebleu:
## Description of the issue/feature this PR addresses

Dovecot is not a hard-dependency of Modoboa, thus it could be installed and running on another host or container - i.e. for isolation reasons. This PR allows one to define - strong - password schemes to use in this case and document the related variables available in the settings.

## Current behavior before PR

If `doveadm` binary is not found on the host, the only password scheme which will be available - and use - will be `{PLAIN}` - which means that password will be stored in clear. There is no alternative to define a *fake* `doveadm` binary - as it is done in tests - to mimic the output and tell Modoboa which schemes are supported - which I think it is not a real and clean solution since Modoboa is at least calling `doveadm` with other arguments.

Password must never be stored in clear, so MD5-CRYPT is added to the fallback list of supported password schemes. Moreover, to handle the case of an installation of Dovecot on another host, a new setting is added to specify which schemes are supported: ``DOVECOT_SUPPORTED_SCHEMES``.

Is it conceivable to add a stronger scheme than MD5-CRYPT by default depending on the [Dovecot documentation](https://doc.dovecot.org/configuration_manual/authentication/password_schemes/)?

### DIFF
--- a/modoboa/core/commands/templates/settings.py.tpl
+++ b/modoboa/core/commands/templates/settings.py.tpl
@@ -114,15 +114,15 @@ MIDDLEWARE = (
 )
 
 AUTHENTICATION_BACKENDS = (
-    # 'modoboa.lib.authbackends.LDAPBackend',
-    # 'modoboa.lib.authbackends.SMTPBackend',
+    #'modoboa.lib.authbackends.LDAPBackend',
+    #'modoboa.lib.authbackends.SMTPBackend',
     'django.contrib.auth.backends.ModelBackend',
 )
 
 # SMTP authentication
-# AUTH_SMTP_SERVER_ADDRESS = 'localhost'
-# AUTH_SMTP_SERVER_PORT = 25
-# AUTH_SMTP_SECURED_MODE = None  # 'ssl' or 'starttls' are accepted
+#AUTH_SMTP_SERVER_ADDRESS = 'localhost'
+#AUTH_SMTP_SERVER_PORT = 25
+#AUTH_SMTP_SECURED_MODE = None  # 'ssl' or 'starttls' are accepted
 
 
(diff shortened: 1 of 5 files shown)


modoboa/core/commands/templates/settings.py.tpl
```diff
@@ -114,15 +114,15 @@ MIDDLEWARE = (
...
     'django.contrib.auth.backends.ModelBackend',
 )
 
 # SMTP authentication
-# AUTH_SMTP_SERVER_ADDRESS = 'localhost'
-# AUTH_SMTP_SERVER_PORT = 25
-# AUTH_SMTP_SECURED_MODE = None  # 'ssl' or 'starttls' are accepted
+#AUTH_SMTP_SERVER_ADDRESS = 'localhost'
+#AUTH_SMTP_SERVER_PORT = 25
+#AUTH_SMTP_SECURED_MODE = None  # 'ssl' or 'starttls' are accepted
```
tonioo, line 125: pep8 says we need a space after a hash comment.

jeromelebleu: It was to distinguish between comments and variables that can be decommented. I can either keep a space or add an exception on this file for this rule to make pep8 happy. What do you prefer?

tonioo: Ok I see. Let's add a exception.

jeromelebleu: I have excluded this error only for this file in the flake8 configuration, is it enough? pep8 seems to be unmaintained for a while, and I didn't find which linter should be run and where.

PR #2147, opened 2021-01-30
Raise domain name length to RFC1035 maximum

sebageek:
RFC1035 Section 2.3.4. specifies a maximum domain name length of 255
octets in protocol. As each label is prefixed by its length and the
domain-root (.) has a length of 0 we have a maximum of 253 octets. For
modoboa to support all these domains in its DB scheme we need to raise
the maximum domain name length to 253 as well.

Modoboa breaks with domain names longer than 100 chars and regretfully I have some domains that are longer than 100 chars.

Current behavior before PR:
Cannot create domain / migration with postfixadm-migrate breaks with exception.

Domain can be successfully created as long as it is <= 253 characters long.


kryskool:
Hi @sebageek 

Thanks for your contribution

It seem correct for me

@tonioo can you check ?

tonioo:
Yes it looks fine, thank you.

kryskool (approved):
Ok for me

PR #2151, opened 2021-02-02
Update Dockerfile.dev

akshatsoni64:
The old version of pip install caused some issues during installation in docker so adding the command to upgrade pip

Description of the issue/feature this PR addresses: This PR will address the issue caused by the old version of the PIP installer

Current behavior before PR: PIP version error

Desired behavior after PR is merged: smooth installation of modoboa server on docker container


Dockerfile.dev
```diff
@@ -13,6 +13,7 @@ RUN python3 -m venv $VIRTUAL_ENV
 WORKDIR /tmp
 COPY requirements.txt /tmp
 COPY test-requirements.txt /tmp
+RUN python -m pip install -U pip
 RUN pip install -r requirements.txt -r test-requirements.txt
```
tonioo, line 17: Could you merge the 2 RUN statements please?

akshatsoni64: It is causing some issues, they need to be kept separated

PR #2159, opened 2021-02-08
Wrap ModelViewSet requests to a revision block

jeromelebleu:
## Description of the issue this PR addresses

Creation, update and deletion of an object is currently not logged when it is done through the API.

## Current behavior before PR

The `create()`, `update()` and `delete()` methods of `rest_framework.viewsets.ModelViewSet` do not create a revision, which do not produce the corresponding log message creation - see `modoboa.core.handlers.post_revision_commit`.

This simply adds a new `modoboa.libs.viewsets.RevisionModelMixin` mixin - sorry, it will produce a conflict with the refacto/vuejs_interface branch - which inherits from [`reversion.views.RevisionMixin`](https://django-reversion.readthedocs.io/en/stable/views.html#reversion-views-revisionmixin). This will decorate the `dispatch()` method to wrap every request in a revision block.

The following can be discussed:

- Is it a good idea to rely on `reversion.views.RevisionMixin` or should Modoboa implements its own mixin?
- Is it okay to decorate all methods - e.g. not just `create()`, `update()` and `delete()`?

### DIFF
--- a/modoboa/lib/viewsets.py
+++ b/modoboa/lib/viewsets.py
@@ -1,0 +1,11 @@
+"""ViewSet related mixins and tools."""
+
+from reversion.views import RevisionMixin
+
+
+class RevisionModelMixin(RevisionMixin):
+    """
+    A mixin to wrap every request in a revision block.
+    """
+
+    pass
--- a/modoboa/admin/api.py
+++ b/modoboa/admin/api.py
@@ -13,6 +13,7 @@ from rest_framework.response import Response
 
 from modoboa.core import models as core_models
 from modoboa.core import sms_backends
+from modoboa.lib.viewsets import RevisionModelMixin
 
 from . import lib, models, serializers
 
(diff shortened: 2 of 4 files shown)


modoboa/lib/viewsets.py
```diff
@@ -0,0 +1,11 @@
+"""ViewSet related mixins and tools."""
+
+from reversion.views import RevisionMixin
+
+
+class RevisionModelMixin(RevisionMixin):
```
tonioo, line 6: Why do you declare this mixin? Why don't you use the RevisionMixin directly?

jeromelebleu: In order to easily update it in a near future, if `RevisionMixin` evolves and does not fit anymore for example. I thought it would prevent to update each viewsets which relies on it - especially in extensions. What do you think?

tonioo: I think it makes sense. Then this PR looks ready to merge right?

jeromelebleu: Yes I think so, also:

> - Is it a good idea to rely on reversion.views.RevisionMixin or should Modoboa implements its own mixin?
> - Is it okay to decorate all methods - e.g. not just create(), update() and delete()?

I think we can see later with time since `modoboa.lib.viewsets.RevisionModelMixin` can easily be modified thanks to that.

tonioo: Yes, I think we can make some tests with what you did and check if it behaves as we want.

PR #2167, opened 2021-02-09
Handle NameTooLong by logging an error

sebageek:
get_dns_records() could be fed with a name that is too long for the DNS
protocol. In this case a NameTooLong error is thrown, which we can catch
and log an error for. This can happen for example if a DNS name that is
already at the limit is used in combination with "manage.py modo
check_mx", which appends an autodiscover to the domain, which then
exceeds the DNS name limit.

`manage.py modo check_mx` crashes each time I run it, errors out with a NameTooLong error.

Current behavior before PR:
Crash

No crash, an error is logged instead.


tonioo:
@sebageek Thank you

Issue #2239, opened 2021-05-07
Active Directory Identity creation/Web login issues


DaveDischord:
Even by looking at the log files, I'm still stumped on where my problem is.  
First, my setup involves modoboa and dovecot syncing with Active Directory.  Dovecot syncing is not through Modoboa's parameter settings.  Most users that have mailboxes set up can access their mail.  Now, however, newly created AD users aren't able to log in Modoboa's web interface. 
 I've ran into the same error logging into Modboa's webmail using 2 different test users:
`**Error: Failed to update password: {'msgtype': 103, 'msgid': 3, 'result': 50, 'desc': 'Insufficient access', 'ctrls': [], 'info': '00000005: SecErr: DSID-031A1256, problem 4003 (INSUFF_ACCESS_RIGHTS), data 0\n'}**`
- One user has an identity/mailbox in Modoboa.  User can create IMAP account under Outlook and Send.Receive emails through that app.
- A freshly created AD user gets the error logging into Modoboa.  No identity or User/Maildir folder is created at this point.  Attempting to create Outlook account, user can log into Incoming email server, but cannot send a test email, leading to a 553 error (Sender address rejected: not owned by user).  At this point, no Identity is created under Modoboa's web interface, but a User/Maildir is created alongside all others.

DaveDischord:
Thanks.  I did update my cron file, but still having the same issue with all test users.  I've ran it manually with the following errors:
**As sudo**
`  File "/srv/modoboa/instance/manage.py", line 16
    ) from exc
         ^
SyntaxError: invalid syntax
`
**In virtual**
After adjusting the search filter under Parameters, it ran without error. 
The Identities section was then populated with other users based on filter, including test users; however, the error message still persists.  

- I've ran an Export on Identities and cells under the column containing the crypted passwords are blank for the test users or any user that has never accessed or hasn't been able to access webmail.

- I've changed all **manage.py** in the cron file except the vmail one to run as modoboa.  All commands executed without error except **manage.py update_statistics** which returned
`rrdtool.OperationalError: opening '/srv/modoboa/rrdfiles/new_accounts.rrd': Permission denied`

- After changes, I've restarted all relevant services and/or restarted the machine completely.

tonioo:
@DaveDischord Your last error seems to be related to an ACL / permission issue on your LDAP server...

im-fabian:
Am I missing something? Because for me it states that the command _import_from_ldap_directory_ does not exist.

```
[...]

PR #2257, opened 2021-06-03
Fixed a bug that was preventing modoboa from properly importing users from an LDAP connection. (modoboa#2239)

ScionOfDesign:
The configuration parameter `ldap_enable_sync` was being queried when it should have been `ldap_enable_import`. (modoboa#2239)
This caused modoboa to attempt to login to the LDAP server with an empty password in an attempt to change said password.

Current behavior before PR:
modoboa would throw an `INVALID_CREDENTIALS ` error from failing to login to the LDAP server with an empty password.

modoboa will successfully update the local password to match the remote LDAP password.

### DIFF
--- a/modoboa/core/app_settings.py
+++ b/modoboa/core/app_settings.py
@@ -18,6 +18,7 @@ from modoboa.parameters import forms as param_forms, tools as param_tools
 from . import constants
 from . import sms_backends
 
+from modoboa.admin.models import Domain
 
 def enabled_applications():
     """Return the list of installed extensions."""
@@ -37,6 +38,18 @@ class GeneralParametersForm(param_forms.AdminParametersForm):
     app = "core"
 
     sep1 = SeparatorField(label=ugettext_lazy("Authentication"))
+
+    domains = [("none", "none")]
+    for domain in Domain.objects.all():
+        domains.append((domain.name, domain.name))
+
+    default_domain = forms.ChoiceField(
+        label=ugettext_lazy("Default domain"),
+        choices=domains,
+        initial="none",
+        help_text=ugettext_lazy(
+            "Default domain to use for accounts.")
+    )
 
     authentication_type = forms.ChoiceField(
         label=ugettext_lazy("Authentication type"),
@@ -526,11 +539,6 @@ class GeneralParametersForm(param_forms.AdminParametersForm):
         self._add_visibilty_rules(
             sms_backends.get_all_backend_visibility_rules()
         )
-        self.fields["password_scheme"].choices = [
-            (hasher.name, ugettext_lazy(hasher.label))
-            for hasher in PasswordHasher.get_password_hashers()
-            if hasher().scheme in get_dovecot_schemes()
-        ]
 
     def _add_dynamic_fields(self):
         new_fields = OrderedDict()
--- a/modoboa/core/models.py
+++ b/modoboa/core/models.py
@@ -117,8 +117,8 @@ class User(AbstractUser):
         :param raw_value: the new password's value
         :param curvalue: the current password (for LDAP authentication)
         """
-        ldap_sync_enable = param_tools.get_global_parameter("ldap_enable_sync")
-        if self.is_local or ldap_sync_enable:
+        ldap_import_enable = param_tools.get_global_parameter("ldap_enable_import")
+        if self.is_local or ldap_import_enable:
             self.password = self._crypt_password(raw_value)
         else:
             if not ldap_available:
--- a/modoboa/core/views/auth.py
+++ b/modoboa/core/views/auth.py
@@ -34,6 +34,8 @@ from .. import sms_backends
 from .. import signals
 from .base import find_nextlocation
 
+import os
+
 logger = logging.getLogger("modoboa.auth")
 
 
@@ -44,11 +46,16 @@ def dologin(request):
         form = forms.LoginForm(request.POST)
         if form.is_valid():
             logger = logging.getLogger("modoboa.auth")
+            if not '@' in form.cleaned_data["username"]:
+                domain = param_tools.get_global_parameter("default_domain",
+                                                          raise_exception=False)
+                if domain is not "none":
+                    form.cleaned_data["username"] += '@' + domain
             user = authenticate(username=form.cleaned_data["username"],
                                 password=form.cleaned_data["password"])
             if user and user.is_active:
                 if param_tools.get_global_parameter("update_scheme",
-                                                    raise_exception=False):
+                                                    raise_exception=False) or not user.is_local:
                     # check if password scheme is correct
                     scheme = param_tools.get_global_parameter(
                         "password_scheme", raise_exception=False)
@@ -64,6 +71,9 @@ def dologin(request):
                         )
                         user.set_password(form.cleaned_data["password"])
                         user.save()
+                    elif not user.is_local:
+                        user.set_password(form.cleaned_data["password"])
+                        user.save()
                     if pwhash.needs_rehash(user.password):
                         logging.info(
                             _("Password hash parameter missmatch. "
@@ -72,7 +82,6 @@ def dologin(request):
                         )
                         user.set_password(form.cleaned_data["password"])
                         user.save()
-
                 login(request, user)
                 if not form.cleaned_data["rememberme"]:
                     request.session.set_expiry(0)
@@ -205,7 +214,7 @@ class VerifySMSCodeView(generic.FormView):
         return HttpResponseRedirect(url)
 
 
-class ResendSMSCodeView(generic.View):
+class ResendSMSCodeView(JSONResponseMixin, generic.View):
     """A view to resend validation code."""
 
     def get(self, request, *args, **kwargs):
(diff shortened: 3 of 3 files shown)


tonioo:
@ScionOfDesign It looks you found a bug but I'm not sure the fix you're providing is the right one... Do you have any python stacktrace to provide? When does this issue happen exactly? The first time you login with an LDAP user? What is your LDAP configuration in Modoboa? Thanks.

tonioo:
@ScionOfDesign I think I understand what is happening! Can you check your configuration to see if the 'Update password scheme at login' setting is enabled? (and I guess you did not enable the 'Enable export to LDAP' option)
If that's the case, here is what is happening:
- Your LDAP user connects to Modoboa, it works
- Modoboa checks if the password scheme needs to be changed
- Since the password field must be empty for this user, Modoboa tries to update the remote password because it is not a local account
And the problem is in the third step, Modoboa should not try to do that for remote accounts.

tonioo:
@ScionOfDesign Could you try the following fix on your platform please?
```diff
diff --git a/modoboa/core/views/auth.py b/modoboa/core/views/auth.py
index 892ec1d2..75a14929 100644
--- a/modoboa/core/views/auth.py
+++ b/modoboa/core/views/auth.py
@@ -47,8 +47,12 @@ def dologin(request):
             user = authenticate(username=form.cleaned_data["username"],
                                 password=form.cleaned_data["password"])
             if user and user.is_active:
-                if param_tools.get_global_parameter("update_scheme",
-                                                    raise_exception=False):
+                condition = (
+                    user.is_local and
+                    param_tools.get_global_parameter(
+                        "update_scheme", raise_exception=False)
+                )
+                if condition:
                     # check if password scheme is correct
                     scheme = param_tools.get_global_parameter(
                         "password_scheme", raise_exception=False)
```

ScionOfDesign:
Here is the stacktrace of the error that occurs when you try to login to an LDAP account _without_ first creating any new user, or if you used the `import_from_ldap_directory` command: https://dpaste.com/8LDAHQHKE

I documented many of the nuances of the bug I discovered in #2239.

I will try out your fix now.

ScionOfDesign:
I tried out your fix.
It works for logging into the Dashboard/Profile Settings and creates the new account properly, but when I try to access webmail I get the same `Error: [b'[AUTHENTICATIONFAILED] Authentication failed.']` that I was getting before when I made the accounts manually.
Also, editing the account in the Identities panel still causes 'Internal Error' to occur.

ScionOfDesign:
For my LDAP server, I am using the FreeIPA Docker image, using a variant of this docker-compose file found [here](https://github.com/Mettmett/docker-compose/tree/master/freeipa).
Here are my LDAP settings:
image

tonioo:
@ScionOfDesign To access the webmail, your username must be a valid email address. This is what dovecot expects.

tonioo:
@ScionOfDesign Do you have a trace for the edit account issue?

ScionOfDesign:
After, applying your fix, I changed the 'username' attribute to 'mail'.
This allowed me to log in using the full email of the user as the username.
After this, the error in the identity panel does not occur, however, I am still _unable_ to log in to the Webmail panel.
 
I do not have a trace for the error in the identity panel, as it was a simple popup. This only seemed to occur with auto-generated accounts that used just the uid as the username.

As a side note, be able to select a 'primary' domain at letting users optionally login with just their uid instead of full email address is something I very much would like. The change I made seemed to allow this, though I am unsure if it would cause unintended consequences.

tonioo:
@ScionOfDesign Make sure the user you're talking has a mailbox defined. (check the Mail tab when you edit it)
When Modoboa imports a user from LDAP, it will be able to create a mailbox only if the username is an email address.

ScionOfDesign:
I tried both users with a full email as well as users with just the username. Both times a local account was created but I could not log into the webmail. When I use the full email, I can edit the user in the identities tab like normal.

With my change, I am able to log in and send emails via webmail with both accounts made with just the username as well as accounts made using the full email. However auto-created accounts with only a username cannot be edited in the identities panel.

Interestingly, if I use the uid as the username and sign into an account, changing the username LDAP attribute to 'mail' does not work. Accounts only accept the username and not the full email. I have to completely recreate the app and then sign in to a new account after reconfiguring the ldap settings.

I would suggest testing out how the sync is working with freeIPA yourself. If you want I could spin up a freeIPA test server for you to use.

ScionOfDesign:
Oh, an update, I was able to find the error log for the internal error that occurs when trying to edit a user created automatically by signing in with _just_ the username: https://dpaste.com/9LTQWQHQA

Interestingly, I can't seem to replicate the behavior of being able to sign in with my LDAP accounts email after I did it once. It only seems to let me sign in with my UID.

tonioo:
@ScionOfDesign Thank you for the trace. According to the screenshot of your settings, Modoboa will always create usernames that are not email addresses (since you use the uid property as a username). This is something that is supposed to work only to create domain administrators in some cases. But as I said, you won't be able to access functionalities that require a mailbox (webmail, calendars, contacts, ...).
Looks like there is a weakness in Modoboa but your setup won't work even if I fix the bug you found. Can't you change the auth method to use the searchbind option? Doing so, you could use the email address directly in the search filter...

ScionOfDesign:
Bind and search seemed to work fine. It also prevented the internal error when trying to edit new users.
Below are some tests with bind and search, clearing the users after each attempt.

Using only my fix:
Search and bind.
Must first use UID to login.
Full email works after, unless you change password.
On a password change, must login with UID.
Cannot login to webmail, must change password locally first to be the same thing.

(Side note, visiting webmail tells you that it requires the state AUTH until you are in a Non-auth state if you first try to visit it, then change your password (which kicks you out) and sign in again to visit webmail. You must log out then log in again to access webmail after updating your local password.)

Using only your fix:
I must use the full email of the user to sign in.
Even still, I cannot sign in to webmail, I get the same authentication error.
Same results after changing passwords on FreeIPA.
But if I try to change my local password I get the error: `Failed to update password: {'msgtype': 103, 'msgid': 3, 'result': 19, 'desc': 'Constraint violation', 'ctrls': [], 'info': 'Pre-Encoded passwords are not valid\n'}`

Using BOTH fixes:
I can only log in with a full email.
Logging in with a full email has the same behavior as your fix only, but this time I can change my local password to fix the problem.

The behavior I want is most replicated by using Bind and Search + My Fix Only. However, it seems that the local password should be set to the LDAP password *every* time a user logs in.
Or it should sync. I have no idea how the `Dovecot Sync` option would affect my situation, as I have not really messed with it and the documentation on it is sparse.

ScionOfDesign:
Alright, I have spun up a FreeIPA server on Vultr. I can keep it up until you're done testing.
Dashboard URL: freeipa.studiowhy.ml
User: admin
Password: modoboamail

LDAP URL: ldap.studiowhy.ml
Port: 389
STARTTLS: No
Bind DN: uid=admin,cn=users,cn=accounts,dc=studiowhy,dc=ml
Bind Password is admin password.

Just create new user accounts and try to have them log in.

ScionOfDesign:
I made some changes to the code to add in the sort of functionality that I'd expect. I'm not a Django developer, so it's not likely merge-worthy but it should give you an idea of what I'd like to see.

The first thing is that it should always update the local password to be the same as the remote password when the user logs in. (Or ideally the local password should *always* be synced.)
The second is that the admin should be able to select a 'default' domain. Users logging in with only their username will have this domain appended onto their username when they log in.
My change would probably break the admin-only functionality you mentioned, but you get the idea.

tonioo:
@ScionOfDesign I've made a test using your FreeIPA instance and indeed it works as expected with search&bind. One thing you must know is that you must always use the same username. For example, if you use an email address the first time, it will be the username of the user created in modoboa and the authentication will always be made against the LDAP server, we'll never store any password locally.
Then for the webmail, we use IMAP and we must authenticate the first time but against dovecot, which is not LDAP ready unless you configure it to (there is an option in Modoboa for that). So, in your case, it's normal if you can't access the webmail I guess...
Your idea to define a default domain for LDAP created users makes sense if you want to user non-email usernames but it will limit your LDAP setup to a single domain and you'll have to update your dovecot configuration accordingly.

tonioo:
@ScionOfDesign I also made a test using direct bind authentication with a simple username and there is indeed a "strange" behaviour in this case. It works but the created account will have username@studiowhy.ml as local username and a studiowhy.ml domain will also be created the first time (due to default settings). To use it, you will always authenticate using a simple username (otherwise the LDAP backend won't find the remote account), not the email address. And you still have to configure dovecot properly if you want to use the webmail. Does it make sense?

ScionOfDesign:
Yes, that makes sense, thank you for looking into it.
I was able to configure my Modoboa instance to work the way I wanted it to (though some more documentation on the dovecot LDAP sync would be appreciated).

My only other issue is the one I posted here: #2262, which I encounter even after removing the second domain.

Also, the ability to sync 'Display Name' with the LDAP server to avoid sending emails with the full real names of users would be appreciated.

Aterfax:
I'm still having to faff around with getting this working with FreeIPA. This may help someone in future? I think the following is working for my Dovecot LDAP config:

```
uris = ldaps://hostname:636
dn = "uid=your_bind_user,cn=sysaccounts,cn=etc,dc=domain,dc=com"
dnpass = 'yourbindpassword'
base = cn=users,cn=accounts,dc=domain,dc=com
auth_bind = yes
auth_bind_userdn = uid=%n,cn=users,cn=accounts,dc=domain,dc=com

user_filter = (&(objectClass=inetOrgPerson)(|(uid=%u)(mail=%u)))
pass_filter = (&(objectClass=inetOrgPerson)(|(uid=%u)(mail=%u)))
```

%u being used twice appears to be needed to keep both the webmail and mail clients (e.g. Thunderbird) happy.

PR #2313, opened 2021-07-05
Fixed some warnings reported by drf-speculator

jeromelebleu:
This fixes the following error and warnings reported by drf-speculator:

```
Error #0: AccountViewSet: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Ignoring view for now.
Warning #1: AccountViewSet: could not resolve authenticator <class 'modoboa.core.drf_authentication.JWTAuthenticationWith2FA'>. There was no OpenApiAuthenticationExtension registered for that class. Try creating one by subclassing it. Ignoring for now.
```

### DIFF
--- a/modoboa/core/drf_authentication.py
+++ b/modoboa/core/drf_authentication.py
@@ -4,6 +4,7 @@ from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
 
 from django_otp.models import Device
+from drf_spectacular.contrib.rest_framework_simplejwt import SimpleJWTScheme
 from rest_framework import status
 from rest_framework.exceptions import APIException
 from rest_framework_simplejwt.authentication import JWTAuthentication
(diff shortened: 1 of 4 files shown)


modoboa/core/drf_authentication.py
```diff
@@ -44,3 +45,8 @@ def authenticate(self, request):
         if user.tfa_enabled:
             self.verify_user(request, user)
         return user, token
+
+
+class SimpleJWTWith2FAScheme(SimpleJWTScheme):
```
tonioo, line 50: Where do you use this class?

jeromelebleu: Nowhere, or at least not from the Modoboa side. From what I understood, it is dynamically added to the drf-speculator's extension registry when the class is defined - somewhere [here](https://github.com/tfranzel/drf-spectacular/blob/0.17.2/drf_spectacular/plumbing.py#L563).

PR #2375, opened 2021-10-12
convert2textfield.py

mrattilakovacs:
convert domain dkim key output to handy format

Current behavior before PR:


tonioo:
@mrattilakovacs Do you plan to complete the PR? Because right now, you're just adding code that is not being used anywhere...

tonioo:
Thanks for the contribution but we prefer to focus on the new UI.

PR #2376, opened 2021-10-12
added convert2textfield function

mrattilakovacs:
dkim key convert into tetxtfield to be able to copy in a handy way

Description of the issue/feature this PR addresses: display dkim key in input text field to easier to copy line by line

Current behavior before PR: dkim key was displayed between quotes, copying was hard.

Desired behavior after PR is merged: easy text copy from input text box


tonioo:
Thanks for the contribution but we prefer to focus on the new UI.

PR #2391, opened 2021-11-02
Fix broken django doc. link in plugin_api.rst

SimonVanacco:
Django's 1.9 documentation is no longer accessible, updating to the updated documentation for Django 2.2, used by modoboa.


tonioo:
@SimonVanacco Thank you

Issue #1612, opened 2018-11-11
Robots.txt is missing from urls.py


JeroenvO:
# Impacted versions

* Modoboa: 1.12.2 and older
* installer used: Yes, but some modifications made
* Webserver: Nginx

# Steps to reproduce
Install modoboa and enable webinterface.

# Current behavior
No robots.txt is defined. Search engines do not now how to index the website. When search engines try to find robots.txt an 404 is raised and the error is mailed to ADMINS (if configured)

# Expected behavior
Robots.txt in urls.py defined, to deny all traffic, as webmail should not be publicly indexed by search engines. Possible fix, add:
`path('robots.txt', lambda r: HttpResponse("User-agent: *\nDisAllow: /", content_type="text/plain"), name='robots')`

# Video/Screenshot link (optional)

PR #2495, opened 2022-04-14
Adding a restrictive robot.txt file

SimonVanacco:
Hi !

This PR addresses the issue #1612 by adding a robots.txt file that denies access to Modoboa's web interface. 

Currently search engines are allowed to index every crawlable modoboa file. This new robots.txt file would prevent any page from being indexed, which is probably a good thing as pointed out by the original issue. 

Please let me know if you think of ways to improve this PR.

### DIFF
--- a/modoboa/core/templates/core/robots.txt
+++ b/modoboa/core/templates/core/robots.txt
@@ -1,0 +1,2 @@
+User-Agent: *
+Disallow: /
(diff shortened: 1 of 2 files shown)


ksaadDE:
Keep in mind it would be better to put Modoboa behind a HTTP Basic Auth to avoid further indexing of Modoboa Sites.
The robots.txt is just a guide, most Search Engines (except Google and Bing) ignoring it on purpose.

PR #2576, opened 2022-07-29
Mitigate spamhaus dns fail while querying for dnsbl

Spitfireap:
Kind of same mitigation as modoboa/modoboa-installer#426 but for dnsbl check. 

I would be more in favor to simply drop spamhaus from dnsbl provider, but this could be a simple mitigation in case the user didn't set a dns resolver with rDNS.

### DIFF
--- a/modoboa/admin/management/commands/subcommands/_mx.py
+++ b/modoboa/admin/management/commands/subcommands/_mx.py
@@ -82,7 +82,7 @@ class CheckMXRecords(BaseCommand):
                 #result from dnsbl is in ipv4 format
                 splited_result = result.split(".")
                 if int(splited_result[-1]) <= 15:
-                    #Typical dnsbl result : 127.255.255.[1-15] (depends on services)
+                    #Typical dnsbl result : 127.0.0.[1-15] (depends on services)
                     result = False
             except socket.gaierror:
                 result = False
--- a/modoboa/admin/management/commands/subcommands/_mx.py
+++ b/modoboa/admin/management/commands/subcommands/_mx.py
@@ -79,6 +79,9 @@ class CheckMXRecords(BaseCommand):
             pattern = "{}.{}.".format(reverse, provider)
             try:
                 result = socket.gethostbyname(pattern)
+                #Mitigate Spamhaus status error (these codes don't mean ip is listed)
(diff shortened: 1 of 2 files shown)


modoboa/admin/management/commands/subcommands/_mx.py
```diff
@@ -79,6 +79,9 @@ def query_dnsbl(self, mx_list, provider):
             pattern = "{}.{}.".format(reverse, provider)
             try:
                 result = socket.gethostbyname(pattern)
+                #Mitigate Spamhaus status error (these codes don't mean ip is listed)
+                if result == "127.255.255.254" or result == "127.255.255.255":
```
tonioo, line 83: I think we should find something more generic to handle that...

Spitfireap: I've tried a more generic way as you proposed, tell me if that's not what you had in mind :)

modoboa/admin/management/commands/subcommands/_mx.py
```diff
@@ -79,6 +79,11 @@ def query_dnsbl(self, mx_list, provider):
             pattern = "{}.{}.".format(reverse, provider)
             try:
                 result = socket.gethostbyname(pattern)
+                #result from dnsbl is in ipv4 format
+                splited_result = result.split(".")
+                if int(splited_result[-1]) <= 15:
+                    #Typical dnsbl result : 127.255.255.[1-15] (depends on services)
+                    result = False
```
tonioo, line 86: That's indeed better like this but if it depends on services then I guess we should store some kind of config somewhere. Do you know if it's documented somewhere?

Spitfireap: It really depends on the provider (spamhaus: https://www.spamhaus.org/faq/section/DNSBL%20Usage#200, spamcop: https://www.spamcop.net/fom-serve/cache/291.html, etc.). When ip is flagged result is 127.0.0.[1-15] and this is share among all providers I checked.

Spitfireap:
I'll check at the tests and fix it later :)

Spitfireap:
@tonioo do you need any other change with this PR ?

PR #2585, opened 2022-08-18
MFA (webauthn, email otp, totp)

Spitfireap:
In progress.


Spitfireap:
So, I'm almost done I think, only to improve render on some page. 

So it's using [django-mfa2](https://github.com/mkalioby/django-mfa2), it is quick and easy, but there are caveats : 
- It won't be compatible with vue interface. --> On this note, vue interface has been here for 2 years it seems and little to no improvement has been done on it. It is only for admin and even there, it lacks parameters... 
- django-mfa2 is managing almost everything so it is not very flexible, although it does work.
- - Pages are not translatable, although it might be possible to see with django-mfa2 devs to implement this.
- implementing a migration for current otp is possible, but perhaps simply removing current OTP implementation is preferable 
- Admin will have to configure manually the "from mail" for email OTP codes sending in settings.py (I don't see a way to use the setting stored in modboa db --> hard coded within django-mfa2).
To sum up, while it sure would be better to have our own django-mfa2 branch, I think it's better to leave it like this, it feels more stable that way 


On this note, the presentation : 
When logged, user will have the MFA page accessible in the top right corner 
Screenshot_20220819_130142

The setup page looks like this : 
Screenshot_20220819_130238

here is an example of FIDO2 key registration : 
Screenshot_20220819_130302

and upon connection, if you have multiple mfa type registered, modoboa will asks to which type you want to use : 
Screenshot_20220819_130123

Regarding personalization, admins can edit settings.py to disable one, or multiple auth type.

Spitfireap:
Currently waiting for upstream recovery code addition.

Issue #2572, opened 2022-07-23
UnicodeEncodeError at /webmail/getmailcontent


jc-plat:
# Impacted versions

* OS Type: Debian
* OS Version: 11.4
* Database Type: MySQL
* Database version: 10.5.15-MariaDB-0+deb11u1
* Modoboa: 20.0.1
* installer used: Yes
* Webserver: Apache

It seems lines 256 &257 in modoboa/lib/email_utils.py are debug trace,
generating an error on some messages.

UnicodeEncodeError at /webmail/getmailcontent
'ascii' codec can't encode character '\xa0' in position 3046: ordinal not in range(128)
Request Method:	GET
Request URL:	http://.../webmail/getmailcontent?mbox=INBOX&mailid=352&links=0
Django Version:	3.2.13
Exception Type:	UnicodeEncodeError
Exception Value:	
'ascii' codec can't encode character '\xa0' in position 3046: ordinal not in range(128)
Exception Location:	/var/modoboa/env/lib/python3.9/site-packages/modoboa/lib/email_utils.py, line 257, in _post_process_html
Python Executable:	/var/modoboa/env/bin/python
Python Version:	3.9.2
Python Path:	
['/var/modoboa/instance',
 '/usr/lib/python39.zip',
 '/usr/lib/python3.9',
 '/usr/lib/python3.9/lib-dynload',
 '/var/modoboa/env/lib/python3.9/site-packages']

Spitfireap:
Hi @jc-plat, do you have an email sample you can share to reproduce the issue ?

jc-plat:
Hi Sitfireap

You will find a sample here: https://ufile.io/dke3tfb1
Simply look this code :
https://github.com/modoboa/modoboa/blob/master/modoboa/lib/email_utils.py#L256
```
        with open("/tmp/output.txt", "w") as fp:
            fp.write(mail_text)
```
html content is stored in /tmp

Spitfireap:
Merci ! So I cannot reproduce the issue. It seems like your system make the default encoding of file to ASCII instead of utf-8. Would you mind editing email_utils.py L256 so it looks like this :
``with open("/tmp/output.txt", "w", encoding="utf-8") as fp:``, basically adding `,encoding="utf-8"`

jc-plat:
It's ok with encoding="utf-8", but remove the dump is better

PR #2590, opened 2022-08-30
removed html dump to file

Spitfireap:
see #2572 

HTML emails would be dumped into ``/tmp/output.txt``. I didn't find any call for this file. Considering it useless.


tonioo:
Good catch, must be some debugging stuff I forgot to remove :)

PR #2598, opened 2022-09-08
Fixed core parameters unable to save on v2

### LINKED ISSUE #2559: Can't save settings in New admin interface (bad request 400)
# Impacted versions
Modoboa 2.0.1

* OS Type: Debian
* OS Version: 11
* Database Type: MySQL
* Database version: ?
* Modoboa: 2.0.1
* Installer used: Yes
* Webserver: Nginx

# Steps to reproduce
1) Login in new interface (mail.example.com/new-admin/) using admin.
2) Go to Settings -> General (/new-admin/parameters/core) and try to save any configuration using the button Save
image

# Current behavior
After clicking the save button. No error message appears and does not save the settings.

Console of my browser:
image

Response:
image

But the toggle "Enable password recovery by SMS" is off:
image

Where is the field "default_top_redirection"???

# Expected behavior
Save the requested configuration

Spitfireap:
fixes #2559 
User can now save core parameters on new-admin.

### DIFF
--- a/modoboa/core/api/v2/serializers.py
+++ b/modoboa/core/api/v2/serializers.py
@@ -107,7 +107,8 @@ class CoreGlobalParametersSerializer(serializers.Serializer):
     )
 
     # Dashboard settings
-    rss_feed_url = serializers.URLField(allow_blank=True, required=False, allow_null=True)
+    rss_feed_url = serializers.URLField(
+        allow_blank=True, required=False, allow_null=True)
     hide_features_widget = serializers.BooleanField(default=False)
 
     # Notification settings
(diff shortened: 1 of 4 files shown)


modoboa/core/api/v2/serializers.py
```diff
@@ -126,7 +127,7 @@ class CoreGlobalParametersSerializer(serializers.Serializer):
     log_maximum_age = serializers.IntegerField(default=365)
     items_per_page = serializers.IntegerField(default=30)
     default_top_redirection = serializers.ChoiceField(
-        default="user", choices=[""])
+        default="user", choices=app_settings.enabled_applications(), required=False)
```
tonioo, line 130: Calling this here is not a good idea because it will be executed at import time and we can't guarantee that all modules will be loaded. You should move this in the serializer constructor.

Spitfireap: Done, thanks

Spitfireap:
Sure, I need to make it automatic on VS code :)

PR #2599, opened 2022-09-13
add python3-certbot-nginx to requirements.txt

ksaadDE:
avoids error messages regarding certbot-nginx py plugin like:
- https://discord.com/channels/930814728928895078/936543858374873138/1019355699337236571 (Modoboa discord, Foghunter message)
- https://github.com/modoboa/modoboa/issues/2129 

Issues with missing python3-certbot-nginx

Error message should not pop up anymore, due to solving of the root cause. 

Note: idk if this is 100% correct, but theoretically it should be.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -38,3 +38,4 @@ qrcode
 
 aiosmtplib
 aioredis==2.0.1
+python3-certbot-nginx # to fix issues like https://discord.com/channels/930814728928895078/936543858374873138/1019355699337236571


Spitfireap:
Added to modoboa/modoboa-installer#428 as well :)

ksaadDE:
@tonioo  @Spitfireap  Well, people will look for a requirements.txt, so they basically need to see it in there. For the installer, I agree it should be there. I don't care if you reject the PR, in case you come to the conclusion it is not needed to put in `requirements.txt`

Spitfireap:
The point of requirements.txt is not to know what is the requirement for modoboa, but it is for pip to install all dependencies modoboa requires. 
Thing is modoboa by itself doesn't require any cert (nginx, dovecot and postfix do). So certbot has nothing to do with modoboa. And it should be included in installer only :D

ksaadDE:
> The point of requirements.txt is not to know what is the requirement for modoboa, but it is for pip to install all dependencies modoboa requires. Thing is modoboa by itself doesn't require any cert (nginx, dovecot and postfix do). So certbot has nothing to do with modoboa. And it should be included in installer only :D

I agree and I would like to only disagree to one point: Visibility, most people will look in modoboa/modoba/requirements.txt instead of the installer. But yes, you are right, if you look at this closer.  To stop that confusion in detail, we could add a requirements.txt in the modoboa-installer repo and load it into the python file, instead of hardcoding it. So we achieve the same thing.

@tonioo  no worries! 👍

Spitfireap:
It's just that : 
1. It is not a requirement of modoboa (what about people running w/o nginx, using dns with certbot, etc.)
2. You need to requirement.txt as a config file and not a readme or something. It is used to really fetch automatically python library *REQUIRED* by modoboa (the command is `pip install -r requirement.txt`.

Adding it in requirement.txt will not achieve the same thing as the installer: if you add it here, it will install certbot-nginx for ALL users, even the one running apache, and it will NOT configure it so basically useless... :)

ksaadDE:
> It's just that :
> 
>     1. It is not a requirement of modoboa (what about people running w/o nginx, using dns with certbot, etc.)
> 
>     2. You need to requirement.txt as a config file and not a readme or something. It is used to really fetch automatically python library _REQUIRED_ by modoboa (the command is `pip install -r requirement.txt`.
> 
> 
> Adding it in requirement.txt will not achieve the same thing as the installer: if you add it here, it will install certbot-nginx for ALL users, even the one running apache, and it will NOT configure it so basically useless... :)

Hello,

no. A requirements.txt contains all the required modules from pip for a Python Application. It is NOT a documation or something like this. 

I was talking about the installer, sorry for the misunderstanding. I asked for a requirements.txt in the installer, instead of hard-coding it into the python code.

Hopefully it is more clear now. 

Regards

Spitfireap:
Oh okay, well If you're not using the installer, then you're doing everything manually (including certificate configuration and renewal). As such, You are entitled to set it up with or without certbot.
I really believe that if you forget/don't know how to do it you shouldn't try to install modoboa manually and furthermore you shouldn't even run a mail server. Don't be like manjaro :p

ksaadDE:
@Spitfireap  I understand that this is all new and rushing to you. But please refrain from personal attacks (it reads a bit like that). 

Most parts of your message were conclusions out of nowhere, so I am asking you, where did they come from? 

> shouldn't even run a mail server
I am sure you can explain what the Mailserver is, of all the Modoboa components ;-)

I don't think you understand my message probably, so I would like to read again and try to understand.

I will say it again: I said we could add a requirements.txt into the modoboa-installer and load this file in the modoboa-installer python file that does the install, so we don't have to hardcode it. It's not that hard to understand 👍

ksaadDE:
> Oh okay, well If you're not using the installer, then you're doing everything manually (including certificate configuration and renewal). As such, You are entitled to set it up with or without certbot. I really believe that if you forget/don't know how to do it you shouldn't try to install modoboa manually and furthermore you shouldn't even run a mail server. Don't be like manjaro :p

No worries friend, we had a little misunderstanding here.

PR #2717, opened 2022-12-21
docs: run bash explicitly for migration

nicosqc:
hope this is sufficient, else let me know if you would change something. Bash won't be a login shell but since we clear the enviroment with ``sudo -i`` this should not have any meaningful impact

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -78,7 +78,7 @@ use them:
 
 .. sourcecode:: bash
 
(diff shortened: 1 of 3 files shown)


Spitfireap:
Thanks, it will help, surely. There are a few others `sudo -i` across the documentation, would you mind editing all of them ? :)

nicosqc:
struggled with my local branch for a bit there (and github closed it therefore) but updated the fork to change all occurrences

PR #2789, opened 2023-02-17
Fix allocated_quota order in v1

moweerkat:

https://github.com/modoboa/modoboa/issues/2771

Current behavior before PR:

Annotation returns None when there is no entities with allocated quotas.

Annotation returns 0 and is correctly displayed.


moweerkat:
Tests are failing with postgres. Can't understand why

PR #2791, opened 2023-02-17
added pdfcredential support for v2

Spitfireap:


frontend/src/components/identities/IdentityList.vue
```diff
@@ -159,18 +201,36 @@ export default {
...
     getMenuItems (item) {
       const result = []
       if (item.type === 'account') {
+        item.possible_actions.forEach(element => {
+          result.push({
+            label: this.$gettext(element.label),
+            icon: element.icon,
+            link: element.url,
+            onClick: () => {
+              window.open(element.url)
```
tonioo, line 216: We should find something more generic here. All actions do not require to open a new window

PR #2885, opened 2023-02-25
Added ability to disable inactive accounts

Spitfireap:


modoboa/core/app_settings.py
```diff
@@ -455,6 +455,13 @@ class GeneralParametersForm(param_forms.AdminParametersForm):
 
     sep3 = SeparatorField(label=ugettext_lazy("Miscellaneous"))
 
+    enable_inactive_accounts = YesNoField(
+        label=_("Enable inactive account tracking"),
+        initial=True,
+        help_text=_("Allow the administrator to set a threshold beyond which "
+                    "an account is considered inactive if the user hasn't logged.")
```
tonioo, line 462: We could maybe specify that threshold is in days?

Spitfireap: It is specified in the actual threshold field though :)

modoboa/core/app_settings.py
```diff
@@ -455,6 +455,14 @@ class GeneralParametersForm(param_forms.AdminParametersForm):
 
     sep3 = SeparatorField(label=ugettext_lazy("Miscellaneous"))
 
+    enable_inactive_accounts = YesNoField(
+        label=_("Enable inactive account tracking"),
+        initial=True,
+        help_text=_("Allow the administrator to set a threshold (in days) "
+                    "beyond which an account is considered inactive "
+                    "if the user hasn't logged")
```
tonioo, line 463: logged in

PR #2894, opened 2023-03-03
Add adsp DNS config to avoid mails enter in spam in hotmail/outlook

blackhold:
Description of the issue/feature this PR addresses: Add adsp DNS config to avoid mails enter in spam in hotmail/outlook

Current behavior before PR: Mails sent through modoboa enters to spam due DNS configuration

Desired behavior after PR is merged: Mails don't enter to spam in hotmail/outlook


wblondel:
Wow, are you sure that ADSP solved the issue you were having? I've never seen any domain with the ADSP DKIM extension.

Its status was even set to "Historic" in 2009 because no one was using it (see https://datatracker.ietf.org/doc/status-change-adsp-rfc5617-to-historic/).

PR #2986, opened 2023-04-28
Migrate to Django 4.2

Spitfireap:


.github/workflows/modoboa.yml
```diff
@@ -97,7 +97,6 @@ jobs:
           pip install -r mysql-requirements.txt
           echo "DB=mysql" >> $GITHUB_ENV
       - name: Test with pytest
-        if: ${{ matrix.python-version != '3.8' || matrix.database != 'postgres' }}
```
tonioo, line 100: Why do you remove this line?

doc/upgrade.rst
```diff
@@ -108,6 +108,23 @@ Rebuild Virtual Environment
 Specific instructions
 *********************
 
+2.1.5
+=====
```
tonioo, line 112: According to modifications, we should maybe increase to 2.2?

.github/workflows/modoboa.yml
tonioo: If you look closely, you'll see that there is one job to run only tests, and one job to run tests and coverage

PR #3014, opened 2023-06-18
Update upgrade.rst imap_migration module fix.

samweisgamdschie:

It's just a `.` rather than a `_`. 

Current behavior before PR:

I copied this line and ran into an 500 Internal server error after login and needed to figure out why via `DEBUG=True` because there where no logs found anywhere pointing to that error.

Logging in is possible again.


tonioo:
@samweisgamdschie Thanks for fixing that.

PR #3026, opened 2023-07-10
Use asyncio from redis instead of aioredis to make app python 3.11 compatible

florealcab:

Policyd part crash on debian 12

Current behavior before PR:

The policyd supervisor crash with a well known bug on aioredis with python 3.11. Like I found [here](https://github.com/aio-libs/aioredis-py/issues/1443), aioredis is not maintained anymore, but we can replace it by redis with asyncio (with redis >= 4.2.0rc1)

I made the change and all seems to work well with python 3.11, but we should want more testers!

### DIFF
--- a/.github/workflows/modoboa.yml
+++ b/.github/workflows/modoboa.yml
@@ -71,7 +71,7 @@ jobs:
     strategy:
       matrix:
         database: ['postgres', 'mysql']
-        python-version: [3.8, 3.9, '3.10']
+        python-version: [3.8, 3.9, '3.10', '3.11']
       fail-fast: false
 
     steps:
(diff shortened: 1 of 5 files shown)


Spitfireap:
Hi, can you add 3.11 version to the github workflow ?

florealcab:
Sure, I add it to the list of python versions in the workflow :)

Spitfireap:
Thanks :). So looks like we need to update to account for the [deprecation](https://docs.python.org/3/whatsnew/3.11.html#removed). Can you look at it ? :D

Spitfireap:
I would advise replacing the coroutine from asynctest with [this](https://docs.python.org/3/library/unittest.mock.html?highlight=asyncmock#unittest.mock.AsyncMock).

florealcab:
Thank you, I use it now. The tests passed I hoped I have set it correctly because I never use it not asynctest before. if I make mistakes, don't hesitate to correct me :)

Spitfireap:
Wonderful, seems good :)

PR #3039, opened 2023-08-04
Improved alias management

Spitfireap:
- Fix Admin not able to add alias with the user creation wizard
- Changed behavior of aliases creation in general : trying to create an already existing alias will simply add the recipient to the existing one silently
- Using continue-if-exists option on alias import from csv will now modify the alias instead of throwing an error (because of bad formatting). Without the option, the error will be thrown
- Improved error message regarding alias creation/modification/importation
- Trying to create an already existing alias on new admin will automatically redirect to the edit page of said alias

### DIFF
--- a/modoboa/admin/models/alias.py
+++ b/modoboa/admin/models/alias.py
@@ -39,7 +39,7 @@ def validate_alias_address(
         condition = (
             alias.exists()
             and not ignore_existing
-            )
+        )
         if condition:
             raise AliasExists(alias.first().pk)
     if instance is None:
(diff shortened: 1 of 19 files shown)


modoboa/admin/models/alias.py
```diff
@@ -49,6 +56,39 @@ def validate_alias_address(address, creator, internal=False, instance=None):
...
+        alias_recipients = list(alias.first().recipients)
+        if recipient_to_delete in alias_recipients:
+            if len(alias_recipients) == 1:
+                # Only recipient, we delete the AliasExists
+                alias.delete()
+            else:
+                alias_recipients.remove(recipient_to_delete)
+                alias = alias.first()
+                alias.set_recipients(alias_recipients)
+                alias.save()
```
tonioo, line 89: I guess this could be replaced by a method in the Alias class directly...

PR #3061, opened 2023-09-11
Few fixes

### LINKED ISSUE #3055: [Bug] Settings don't load on new-admin after reloading the page
On modoboa 2.2.0, if the user reloads the page, it seems that the parameters.getApplications() never gets called and users have no option for the settings drawer.

Spitfireap:
- Fixed pdfcredentials preventing login
- Fixed message_limit issue on domains
- Improved Resources and Role selection for accounts
- Better quota displayed when not restricted
- Fixed settings option not loading when reloading the page (Fixes #3055)
- Improved dkim management and displaying

### DIFF
--- a/frontend/src/components/identities/AccountCreationForm.vue
+++ b/frontend/src/components/identities/AccountCreationForm.vue
@@ -11,7 +11,7 @@
   @create="submit"
   >
   <template v-slot:form.role="{ step }">
-    <account-role-form :ref="`form_${step}`" v-model="account.role" :account="account" />
+    <account-role-form :ref="`form_${step}`" v-model="account.role" :account="account" :authUserRole="authUser.role" />
   </template>
   <template v-slot:form.identification="{ step }">
     <account-general-form v-if="step >= 2" :ref="`form_${step}`" v-model="account" />
@@ -50,6 +50,7 @@
 </template>
 
 <script>
+import { mapGetters } from 'vuex'
 import { bus } from '@/main'
 import accounts from '@/api/accounts'
 import AccountAliasForm from './AccountAliasForm'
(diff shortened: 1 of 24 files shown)


frontend/src/components/identities/AccountCreationForm.vue
```diff
@@ -11,7 +11,7 @@
   @create="submit"
   >
   <template v-slot:form.role="{ step }">
-    <account-role-form :ref="`form_${step}`" v-model="account.role" :account="account" />
+    <account-role-form :ref="`form_${step}`" v-model="account.role" :account="account" :authUserRole="authUser.role" />
```
tonioo, line 14: Since there is only one authUser and it comes from a store, maybe you could use the getter directly inside the account-role-form component?

frontend/src/components/identities/AccountForm.vue
```diff
@@ -27,7 +27,7 @@
         </v-row>
       </v-expansion-panel-header>
       <v-expansion-panel-content>
-        <account-role-form ref="roleForm" v-model="editedAccount.role" :account="editedAccount" />
+        <account-role-form ref="roleForm" v-model="editedAccount.role" :account="editedAccount" :authUserRole="authUser.role"/>
```
tonioo, line 30: Since there is only one authUser and it comes from a store, maybe you could use the getter directly inside the account-role-form component?

frontend/src/components/identities/AccountForm.vue
```diff
@@ -137,15 +137,37 @@
...
+            <v-fade-transition leave-absolute>
+              <span v-if="open"></span>
+              <v-row
+                v-else
+                no-gutters
+                style="width: 100%"
+                >
+                <template v-if="account.resources != undefined && account.resources.length == 2">
+                  <v-col cols="6">
+                    <translate class="mr-2">Mailbox: </translate> <translate>Number of allowed mailbox: </translate> {{ account.resources[0].max_value}}
```
tonioo, line 158: allowed mailboxes

Issue #3165, opened 2024-01-12
dns check should be made against a domain authoritative server


arthru:
Dns checks are made using the local dns resolver or a configured one.

By using a non authoritative resolver, we have to wait for dns propagation to be made, that could be long depending to the configured TTL.

I propose to use the dns resolver to find the authoritative server and then to make dns request against that authoritative server.

With that change, dns checks would be instantaneous

PR #3166, opened 2024-01-12
use authoritative server to perform dns checks

arthru:
Description of the issue/feature this PR addresses: see #3165

### DIFF
--- a/modoboa/admin/lib.py
+++ b/modoboa/admin/lib.py
@@ -199,8 +199,15 @@ def get_authoritative_server(domain, resolver=None):
     while True:
         try:
             answers = dns.resolver.resolve(dnsname, 'NS')
-            return str(answers[0].target)
+            if answers:
+                first_answer = answers[0]
+                if hasattr(first_answer, "target"):
+                    return str(first_answer.target)
+                if hasattr(first_answer, "address"):
+                    return str(first_answer.address)
         except dns.resolver.NoAnswer as e:
+            dnsname = dnsname.parent()
+        else:
             dnsname = dnsname.parent()
 
 
--- a/modoboa/admin/lib.py
+++ b/modoboa/admin/lib.py
@@ -190,11 +190,45 @@ def get_dns_resolver():
     return resolver
 
 
+def get_authoritative_server(domain, resolver=None):
+    if not resolver:
+       resolver = get_dns_resolver()
+
+    dnsname = dns.name.from_text(domain)
+
+    while True:
+        try:
+            answers = dns.resolver.resolve(dnsname, 'NS')
+            return str(answers[0].target)
+        except dns.resolver.NoAnswer as e:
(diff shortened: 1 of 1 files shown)


Spitfireap:
Hi, any news on this PR @arthru ?

arthru:
Hi

I had not any time at the moment to make the tests pass but thank to your message @Spitfireap , I took the hour needed to check it...

The behavior of asking the authoritative server of example.com was not the same than with my resolver when I wrote the code  so I had to enhance my code.

I just added a commit and rebased against master so I hope the tests are going to pass now. Let'em run !

modoboa/admin/lib.py
```diff
@@ -190,11 +190,52 @@ def get_dns_resolver():
...
+            if answers:
+                first_answer = answers[0]
+                if hasattr(first_answer, "target"):
+                    return str(first_answer.target)
+                if hasattr(first_answer, "address"):
+                    return str(first_answer.address)
+        except dns.resolver.NoAnswer as e:
+            dnsname = dnsname.parent()
+        else:
+            dnsname = dnsname.parent()
```
tonioo, line 211: @arthru What happens if parent is None?

arthru: the ultimate parent is `.` that is not shown on most domains

A ns query on `.` answers to the root dns servers that always answers :

```
$ dig ns .

; <<>> DiG 9.18.27 <<>> ns .
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 58155
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 13, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1472
;; QUESTION SECTION:
;.				IN	NS

;; ANSWER SECTION:
.			74393	IN	NS	a.root-servers.net.
.			74393	IN	NS	b.root-servers.net.
.			74393	IN	NS	c.root-servers.net.
.			74393	IN	NS	d.root-servers.net.
.			74393	IN	NS	e.root-servers.net.
.			74393	IN	NS	f.root-servers.net.
.			74393	IN	NS	g.root-servers.net.
.			74393	IN	NS	h.root-servers.net.
.			74393	IN	NS	i.root-servers.net.
.			74393	IN	NS	j.root-servers.net.
.			74393	IN	NS	k.root-servers.net.
.			74393	IN	NS	l.root-servers.net.
.			74393	IN	NS	m.root-servers.net.

;; Query time: 10 msec
;; SERVER: 192.168.111.254#53(192.168.111.254) (UDP)
;; WHEN: Wed Jul 17 14:55:13 CEST 2024
;; MSG SIZE  rcvd: 239
```

so the parent should never be None

tonioo: Ok I see.

PR #3199, opened 2024-02-21
Updated translations on frontend

Spitfireap:
Updated CI job as well


frontend/src/plugins/gettext.js
```diff
@@ -16,6 +16,7 @@ export default createGettext({
     nl: 'Dutch',
     pl: 'Polish',
     pt: 'Portuguese',
+    'pt-br': 'Brazilian Protuguese',
```
tonioo, line 19: Portuguese

Issue #2331, opened 2021-07-20
Feature Request: WebAuthn


10Meisterbaelle:
WebAuthn support for hardware-based authentication

meramsey:
Something like this added would be awesome
https://github.com/mkalioby/django-mfa2

meramsey:
still no interest in this?

bcsanford:
not sure how it was done but Mesh Central Supports this works great with my Yubikey, would love to have it with Modoboa

Spitfireap:
I'm working on it, see #2585 to see advancement :)

rivarob04:
I am highly intrested in this. How do I get more people on board?

PR #3217, opened 2024-03-27
sieve filters & webauthn

tonioo:


docker/.Dockerfile.front.dev.kate-swp
tonioo, line 1: I think this file is useless

Spitfireap:
closes #2331

PR #3247, opened 2024-05-05
Documenting use of Redis socket and fixing configuration

### LINKED ISSUE #3241: Documentation lacks using unix socket
# Impacted versions

* OS Type: Arch
* OS Version: current
* Database Type: MySQL
* Database version: 11.3.2
* Modoboa: 2.2.4
* installer used: No
* Webserver: Nginx

The documentation assumes that all connections to external services use TCP, but unix socket may be used too.

I struggled to configure MySQL this way since it's not documented, but now that redis is required, I can't figure out how to use it.

I use this in settings.py:

```
REDIS_HOST = '/run/redis/redis.sock'
REDIS_PORT = 6279
REDIS_QUOTA_DB = 0
REDIS_URL = 'unix://{}?db={}'.format(REDIS_HOST, REDIS_QUOTA_DB)
```

But django seems to keep using REDIS_PORT as i get this error:

```
ConnectionError at /

Error -2 connecting to /run/redis/redis.sock:6279. Name or service not known.
```

Overwriting REDIS_PORT with 0 or empty value keeps throwing errors.

So, how to use redis unix socket?

yannfill:
Fixes #3241.

Current behavior before PR:
Can't get Modoboa working with Redis socket.

Using Redis socket is fixed in `settings.py` and documented.

### DIFF
--- a/doc/upgrade.rst
+++ b/doc/upgrade.rst
@@ -766,7 +766,7 @@ Add the following settings to your ``settings.py`` file:
    REDIS_QUOTA_DB = 0
    REDIS_URL = 'redis://{}:{}/{}'.format(REDIS_HOST, REDIS_PORT, REDIS_QUOTA_DB)
 
-Or, if Redis listen on unix socket:
+Or, if Redis listen on unix socket and you are not using the Modoboa installer:
 
 .. sourcecode:: python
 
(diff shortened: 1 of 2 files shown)


doc/upgrade.rst
Spitfireap, line 1: Could you explain that this configuration is not to be used when using the installer ?

yannfill: It's done

Spitfireap:
Nice, thanks for the contribution

PR #3251, opened 2024-05-14
Update login.html

RJohnPaul:
Changes made as reqested with Optimized CSS Loading ,  Disabled submit button on form submission to prevent multiple submissions ,  Added aria-label attributes for better accessibility.


tonioo:
@RJohnPaul We updated login page with new CSS and layout so unfortunately this PR has become obsolete. Thanks anyway :)

Issue #3327, opened 2024-10-25
Setting a password policy that requires "special" symbols breaks the random password generator


rlKoekie:
Modoboa 2.2.4 on Ubuntu 22.04.4
Priority: low

Backstory: 
I stumbled over AUTH_PASSWORD_VALIDATORS in /srv/modoboa/instance/instance/settings.py , and decided to require passwords to have a special character in them. This has the unfortunate side effect of throwing the random password generator in an endless loop, causing the web interface to become unresponsive. 
This happens because the make_password function in /srv/modoboa/env/lib/python3.10/site-packages/modoboa/admin/lib.py generates a random password consisting of only letters and numbers, then checks if it is valid, and if not it will generate a new one. It keeps looping until it generates a valid password, which of course is never if the validator requires it to contain a special character.

Steps to reproduce:
* Set   'specials': 1   in  /srv/modoboa/instance/instance/settings.py
* Create a new account, and leave the random password toggle enabled
* Hit "next", the interface will freeze until you call "systemctl restart uwsgi.service" on the box.

Possible fix:
I understand that this is completely my own doing by changing the settings.py, and that this is most likely an unsupported setup. I mostly wanted to document this here for the next overzealous person to trip over this. If you are this person: you can patch the make_password function in /srv/modoboa/env/lib/python3.10/site-packages/modoboa/admin/lib.py to insert special characters as well:
```
def make_password():
    """Create a random password."""
    length = int(
        param_tools.get_global_parameter("random_password_length", app="core")
    )
    while True:
        password = "".join(
            random.SystemRandom().choice(
                # HvS: add string.punctuation here to generate valid passwords
                string.ascii_letters + string.digits + string.punctuation) for _ in range(length))
        try:
            password_validation.validate_password(password)
        except ValidationError:
            continue
        return password
```
It might also be nice to have a maximum number of tries in this loop, to prevent it from getting stuck. But we leave that as an exercise for the reader :-)

rlKoekie:
small sidenote: the fix I posted above is probably not the way to go for everybody, as it generates quite ugly passwords with many special characters. I got presented with this beauty on my first attempt: 3@M|@5t@Tyv!

Spitfireap:
Could you maybe post a PR ? 
I think we should add a toggle for special characters (and provide a list of possible characters). The loop should also be limited you're right

PR #3333, opened 2024-11-01
Add YesorNo Field option for a special char in random password generation, also make sure that special char always present if intended

FuzzyParrabellum:
cf issue #3327 - Setting a password policy that requires "special" symbols breaks the random password generator

Before PR, if user wanted to have a special character in their randomly generated password upon user creation, they had to directely change AUTH_PASSWORD_VALIDATORS in settings.py. That behavior could lead to an infinite while loop bug with custom make_password() function in modoboa\modoboa\admin\lib.py.

After PR :

- User directly has option (with yes/no radiobutton) in its account parameters to **allow for special characters** in new randomly generated user passwords.
-  Now random password generator function make_password() always include at least one of the possible string groups **ascii_letters**, **digits** (and **punctuation**/**special characters**, if user chose so in its account parameters).
- Set a max of **10 tries** of make_password() call.

### DIFF
--- a/modoboa/admin/lib.py
+++ b/modoboa/admin/lib.py
@@ -325,16 +325,13 @@ def make_password():
         for possible_chars in possible_chars_types:
             all_possible_chars += possible_chars
             certain_chars += random.SystemRandom().choice(possible_chars)
-        
         password = "".join(
             random.SystemRandom().choice(all_possible_chars) \
                 for _ in range(length - len(possible_chars_types))
         ) + certain_chars
-        
         li_password = list(password)
         random.shuffle(li_password)
         password = "".join(li_password)
-        
         try:
             password_validation.validate_password(password)
         except ValidationError:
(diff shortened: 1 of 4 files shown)


modoboa/admin/lib.py
```diff
@@ -313,11 +313,25 @@ def domain_has_authorized_mx(name):
...
     """Create a random password."""
     length = int(param_tools.get_global_parameter("random_password_length", app="core"))
-    while True:
+    allow_special_characters = bool(param_tools.get_global_parameter("allow_special_characters", app="core"))
+    possible_chars_types = [string.ascii_letters, string.digits]
+    all_possible_chars = ''
+    certain_chars = ''
+    for i in range(10): # limit tries
+        if allow_special_characters: # add special characters in random passwd
+            possible_chars_types.append(string.punctuation)
```
Spitfireap, line 322: Aren't you adding it at each try ?

modoboa/admin/tests/test_account.py
```diff
@@ -346,6 +346,10 @@ def test_random_password(self):
             "email": "tester@test.com",
             "stepid": "step2",
         }
+        # TODO: add test to check if unhashed password created with
```
Spitfireap, line 349: Have you got time to do it ?

Spitfireap (dismissed):

PR #3342, opened 2024-11-14
Update djangojs.po

slaweallx:
I translated the entire language file into Turkish from start to finish with the finest detail. I compared it with the Comparex tool to avoid errors.
I made the whole language file clean, please review it.


tonioo:
@slaweallx Thank you!

PR #3456, opened 2025-02-13
Only symlink to frontend directory contents instead of copying them

ntninja:
Makes `load_initial_data` only create symlinks to frontend directories rather than (unnecessarily) copying them. This also avoids problems with Python’s `shutil.copytree` copying permission bits when copying from a Modoboa source on a read-only filesystem (as is the case on NixOS), which then causes the call to `shutil.rmtree` to fail on subsequent invocations of the `load_initial_data` command.

Current behavior before PR:
Frontend files are copied.

Frontend files are no longer copied, but rather just symlinked. Only the “config.json” file is locally created based on Modoboa configuration.


Spitfireap:
Hi, thanks for the contribution, it seems cleaner indeed. Do you have some time [to add a test around here](https://github.com/modoboa/modoboa/blob/df9679a7634b67bf73c54c0251743f67ea19225b/modoboa/core/tests/test_core.py#L63) to check for the symlinks and config.json ?
@tonioo I think we need to investigate about the environment variable for the gh action :p.

ntninja:
@Spitfireap Added a test and made it work

Spitfireap:
Looks good to me (@tonioo).
@ntninja cloud you rebase your branches ounce #3489 gets merges ?

ntninja:
@Spitfireap Passes now after changes done upstream!

tonioo:
@ntninja Can I ask why you create multiple symlinks instead of one global one? Is it because of the config.json file?

ntninja:
@tonioo: Yes, it’s so that we don’t end up writing to the symlink target which may reside on a read-only filesystem (in NixOS at least the Nix store – where Modoboa data files will be – is only remounted read-write during package operations and left read-only otherwise).

PR #3457, opened 2025-02-13
Prevent numbers used in templates as part of JavaScript or CSS from locale-aware formatting applied if `USE_THOUSANDS_SEPARATOR` is True

ntninja:
Currently setting `USE_THOUSANDS_SEPARATOR = True` in “settings.py” causes the calendar view of modoboa-radicale to break since it expects the global `userLang` value to be set, but the block containing this value fails to parse as valid JavaScript due to a local-aware number being inserted:

```javascript
      var root_url = "/".replace(/\/$/, "");
      var top_notifications;
      var userLang = "de";

      set_static_url("/sitestatic/");
      $(document).ready(function() {
          top_notifications = new TopNotifications({
              url: "/core/top_notifications/check/",
              interval: 30'000  // ← Doesn’t parse!
          });
      });
```

I tried to review and fix all other templates that seem to be affected (only inline CSS rules that wouldn’t cause any major breakage if the failed to parse), but there might be omissions…

Current behavior before PR:
Setting `USE_THOUSANDS_SEPARATOR = True` in “settings.py” caused subtle breakage and the modoboa-radicale view to not render.

That issue no longer applies.

### DIFF
--- a/modoboa/templates/common/buttons_list.html
+++ b/modoboa/templates/common/buttons_list.html
@@ -9,14 +9,14 @@
       {% if entry.menu %}
         {% if entry.url %}
       <a name="{{ entry.name }}" class="btn {{ entry.class }}" href="{{ entry.url }}" title="{{ entry.title }}">
-        {% if entry.img %}<span class="{{ entry.img}}"></span> {% endif %}{{ entry.label }}</a>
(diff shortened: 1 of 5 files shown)


tonioo:
Legacy interface has been removed, this PR is not useful anymore.
But thanks for your contribution!

PR #3490, opened 2025-03-03
Import Rspamd to main repo

Spitfireap:
@tonioo I've got [an issue with this](https://github.com/modoboa/modoboa/blob/bd2e5c76c3604392fc6bc9b472874e83283d7704/frontend/src/layouts/admin/AdminLayout.vue#L99). Any idea on how to make the js object reactive to a ref() ? Else I can drop the feature.

### DIFF
--- a/frontend/src/api/capabilities.js
+++ b/frontend/src/api/capabilities.js
@@ -1,0 +1,7 @@
+import repository from './repository'
+
+export default {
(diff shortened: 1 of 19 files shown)


tonioo:
@Spitfireap I pushed a fix attempt. Can you validate on your side?

PR #3593, opened 2025-07-17
Some fix for 2.4.1

Spitfireap:
- Added favicon for the login page
- Added checks for django and radicale oauth app
- Forced DRF to >=3.16.0 (compatibility with Django 5.2)
- Some doc update


doc/upgrade.rst
```diff
@@ -217,14 +200,20 @@ Required changes to :file:`settings.py`
...
 
-- Remove ``ckeditor`` related apps from ``INSTALLED_APPS`` variable
+- Remove ``ckeditor`` related apps from ``INSTALLED_APPS`` variable and ``CKEDITOR_*`` settings
 
 - Remove ``STATICFILE_DIRS`` variable
 
+Radicale configuration changes
+-----------------------------
+
+You must enable oauth2 authentication in Radicale to make the calendard
```
tonioo, line 214: calendard -> calendars

modoboa/core/management/commands/load_initial_data.py
```diff
@@ -126,18 +129,18 @@ def handle(self, *args, **options):
             call_command(
                 "createapplication",
                 "--algorithm=RS256",
-                f"--redirect-uris={redirect_uri}",
+                f"--redirect-uris={redirect_uris}",
```
tonioo, line 132: Have you checked if it works well with more than one uri? (since you're using space a separator, I have some doubts...)

Spitfireap: Not tested tbf but shows like it in the [doc](https://github.com/jazzband/django-oauth-toolkit/blob/d862f531a2ab96fb3cf05f6ffbf9baf8192c1459/oauth2_provider/management/commands/createapplication.py#L35)

Issue #3655, opened 2025-08-04
Registering webauthn device not working


mbaechtold:
<img width="1512" height="944" alt="Image" src="https://github.com/user-attachments/assets/070eb7e8-1cca-40ea-91d1-ad922dda6941" />

Spitfireap:
So this is through docker ?

mbaechtold:
The backend is running in PyCharm. All other Services are running in Docker.

Spitfireap:
Ok, so I see what's happening : https://bugzilla.mozilla.org/show_bug.cgi?id=1975630
I don't have any idea on how to circumvent it. I mean, it doesn't seem clean to import new CAs for the dev SSL certs...

mbaechtold:
Ok, thank you. Feel free to close the issue.

Spitfireap:
np, see #3672, I have improved the form and you can make it work with chromium locally if you want to test it (you just have to first trust https://localhost:8000 cert or [allow any cert](https://mswjs.io/docs/recipes/using-local-https#chrome--brave) then trust the frontend one https://localhost:3000)

Spitfireap:
Upstream issue : https://bugzilla.mozilla.org/show_bug.cgi?id=1975630
Workaround : Use chromium locally to test the fido feature
Long term fix : perhaps make the certs trusted on the dev environment ?

PR #3672, opened 2025-08-08
Better handling of webauthn + removed deprecated extension

Spitfireap:
See #3655


Spitfireap:
Note : not relying on https://github.com/github/webauthn-json because it has gone into deprecation.
This means relying on this implementation : https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/parseCreationOptionsFromJSON_static which is fairly new, idk if we keep the deprecated repo for now ?

PR #3677, opened 2025-08-13
Allow policy daemon to listen on Unix domain socket and systemd socket activation

ntninja:

1. Allowing policy daemon to listen on Unix domain socket means no port conflicts and less attack surface (since only processes with appropriate permissions may connect to the socket)
2. systemd socket activation means policy daemon is only started on first received message rather than directly on boot

Current behavior before PR:
Policy daemon must be started on boot and bind on loopback TCP

Both limitations are lifted

### DIFF
--- a/modoboa/policyd/management/commands/policy_daemon.py
+++ b/modoboa/policyd/management/commands/policy_daemon.py
@@ -27,6 +27,7 @@ class Command(BaseCommand):
         """Add command line arguments."""
         parser.add_argument("--host", type=str, default="localhost")
         parser.add_argument("--port", type=int, default=9999)
+        parser.add_argument("--socket", type=str, default=None)
         parser.add_argument("--debug", action="store_true", help="Enable debug mode")
 
     def handle(self, *args, **options):
@@ -36,9 +37,14 @@ class Command(BaseCommand):
         except RuntimeError:
             loop = asyncio.new_event_loop()
             asyncio.set_event_loop(loop)
-        coro = asyncio.start_server(
-            core.new_connection, options["host"], options["port"]
-        )
+        if options["socket"] is None:
+            coro = asyncio.start_server(
+                core.new_connection, options["host"], options["port"]
+            )
+        else:
+            coro = asyncio.start_unix_server(
+                core.new_connection, options["socket"]
+            )
         server = loop.run_until_complete(coro)
 
         # Schedule reset task
@@ -59,8 +65,9 @@ class Command(BaseCommand):
 
         logger.info("Stopping policy daemon...")
         # Close the server
-        server.close()
-        loop.run_until_complete(server.wait_closed())
+        for s in server:
+            s.close()
+            loop.run_until_complete(s.wait_closed())
         # Cancel pending tasks
         for task in asyncio.all_tasks(loop):
             task.cancel()
--- a/modoboa/policyd/management/commands/policy_daemon.py
+++ b/modoboa/policyd/management/commands/policy_daemon.py
@@ -36,6 +36,9 @@ class Command(BaseCommand):
         """Entry point."""
         try:
             loop = asyncio.get_event_loop()
+            if loop.is_closed():
+                # Raised on Python 3.11- after fork when *using* the loop
+                raise RuntimeError("Event loop is closed")
         except RuntimeError:
             loop = asyncio.new_event_loop()
             asyncio.set_event_loop(loop)
--- a/modoboa/policyd/management/commands/policy_daemon.py
+++ b/modoboa/policyd/management/commands/policy_daemon.py
@@ -5,6 +5,8 @@ from contextlib import suppress
 import functools
 import logging
 import signal
+import socket
+import os
 
 from django.core.management.base import BaseCommand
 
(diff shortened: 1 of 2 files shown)


modoboa/policyd/management/commands/policy_daemon.py
```diff
@@ -27,14 +29,35 @@ def add_arguments(self, parser):
...
+        parser.add_argument("--socket", type=str, default=None)
         parser.add_argument("--debug", action="store_true", help="Enable debug mode")
 
     def handle(self, *args, **options):
         """Entry point."""
         loop = asyncio.get_event_loop()
-        coro = asyncio.start_server(
-            core.new_connection, options["host"], options["port"]
-        )
+        if os.environ.get("LISTEN_PID", "") == str(os.getpid()) and os.environ.get("LISTEN_FDS", "").isnumeric():
```
tonioo, line 38: Why do you rely on those environment variables? Could you use command line arguments instead?

tonioo: And actually, could you please explain what those variables mean?

ntninja: That is a short implementation of systemd socket activation in Python

For a background see:
 * https://0pointer.de/blog/projects/socket-activation.html (developer tutorial)
 * https://www.man7.org/linux/man-pages/man3/sd_listen_fds.3.html (equivalent function in `libsystemd`)

`sd_listen_fds` is defined as:

```c
#define SD_LISTEN_FDS_START 3
int sd_listen_fds(int unset_environment) {
        int r, fd;
        const char *e;
        char *p = NULL;
        unsigned long l;

        if (!(e = getenv("LISTEN_PID"))) {
                r = 0;
                goto finish;
        }

        errno = 0;
        l = strtoul(e, &p, 10);

        if (errno != 0) {
                r = -errno;
                goto finish;
        }

        if (!p || *p || l <= 0) {
                r = -EINVAL;
                goto finish;
        }

        /* Is this for us? */
        if (getpid() != (pid_t) l) {
                r = 0;
                goto finish;
        }

        if (!(e = getenv("LISTEN_FDS"))) {
                r = 0;
                goto finish;
        }

        errno = 0;
        l = strtoul(e, &p, 10);

        if (errno != 0) {
                r = -errno;
                goto finish;
        }

        if (!p || *p) {
                r = -EINVAL;
                goto finish;
        }

        for (fd = SD_LISTEN_FDS_START; fd < SD_LISTEN_FDS_START + (int) l; fd ++) {
                int flags;

                if ((flags = fcntl(fd, F_GETFD)) < 0) {
                        r = -errno;
                        goto finish;
                }

                if (flags & FD_CLOEXEC)
                        continue;

                if (fcntl(fd, F_SETFD, flags | FD_CLOEXEC) < 0) {
                        r = -errno;
                        goto finish;
                }
        }

        r = (int) l;

finish:
        if (unset_environment) {
                unsetenv("LISTEN_PID");
                unsetenv("LISTEN_FDS");
        }

        return r;
}
```

The code in question reimplements this one function in Python, taking care to:

  * Ensure the listen FDs are intended for our process, rather than the parent (safety check)
  * Not inherit the listen FDs to child processes after exec (O_CLOEXEC)
  * Unset the environment variables after use

The environment variables mean:

  * `LISTEN_PID`: The PID of the process expected to perform socket activation
  * `LISTEN_FDS`: The number of file descriptors inherited to the target process (starting at 3)

tonioo: Ok I see. Could you cover this with some tests?

ntninja: @tonioo: So inline implementation is fine, then? (Just checking to make sure.)

ntninja:
We don’t have much time unfortunately (ME/CFS illness), will get to it 
eventually…


Am Donnerstag, 13. November 2025, 16:45:45 Mitteleuropäische Normalzeit 
schrieb Antoine Nguyen:
> @tonioo commented on this pull request.
> 
> >          parser.add_argument("--debug", action="store_true", help="Enable
> >          debug mode")
>      def handle(self, *args, **options):
>          """Entry point."""
>          loop = asyncio.get_event_loop()
> -        coro = asyncio.start_server(
> -            core.new_connection, options["host"], options["port"]
> -        )
> +        if os.environ.get("LISTEN_PID", "") == str(os.getpid()) and
> os.environ.get("LISTEN_FDS", "").isnumeric():
> 
> @ntninja Any news about tests?

ntninja:
Still planning on doing this, just forgot.

ntninja:
Well, I finally added tests 🥵 (tough)

For some opaque reason it seems to fail on CPython 3.10 and 3.11, I have no idea what the problem is there, but it does not seem to be related to the test.

tonioo:
@ntninja Tests are failing under Python 3.10 and 3.11. Could you check why please?

ntninja:
That worked!? Alright, then! No one knows why it was broken in Python 3.11- anyways! 👀

ntninja:
I don’t think that detected coverage for the `subprocess.Popen`-based subprocess… Please advise!

https://app.codecov.io/gh/modoboa/modoboa/pull/3677?dropdown=coverage&src=pr&el=h1

test_project/test_project/settings.py
```diff
@@ -327,8 +327,8 @@
 
 CACHES = {
     "default": {
-        "BACKEND": "django.core.cache.backends.redis.RedisCache",
-        "LOCATION": f"redis://{REDIS_HOST}:{REDIS_PORT}",
+        "BACKEND": "django.core.cache.backends.filebased.FileBasedCache",
```
kryskool, line 330: @tonioo do you need to change cache backend in this PR ?

ntninja: No, I can put it in a different PR too. Are you opposed to this change in general?

kryskool: Hi @ntninja 

We prefer to have consistent PRs; if a change is unrelated to the PR's objective, it should be moved to another PR, or a ticket should be created to explain the problem.

Regards,

ntninja: Didn’t answer the question, but I moved the commit: #4051

PR #3679, opened 2025-08-15
Update MessageList.vue

PetarPetrushev:
Make it so that on hover of sender address in admin message list log it shows a tooltip of the full untruncated sender email, but still keeps the truncation in the ui.

Description of the issue/feature this PR addresses: Currently if in the admin message list ui there is a long sender email the user only sees the truncated email and has no way to view full email. 

Current behavior before PR: Currently if in the admin message list ui there is a long sender email the user only sees the truncated email and has no way to view full email. 

Desired behavior after PR is merged: Now if the user hovers over the truncated email their browser will show a tooltip showing the full sender email while still keeping the sender email truncated in the ui.

### DIFF
--- a/frontend/src/components/admin/logs/MessageList.vue
+++ b/frontend/src/components/admin/logs/MessageList.vue
@@ -25,7 +25,9 @@
         {{ $date(item.date) }}
       </template>
       <template #[`item.sender`]="{ item }">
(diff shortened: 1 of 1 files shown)


tonioo:
@PetarPetrushev Nice catch, thanks.

PR #3750, opened 2025-10-13
Stop logging DNS issues to the database

tonioo:
This creates too many entries for nothing really useful.


mirtouf:
I confirm this is should be optional:
image

PR #3765, opened 2025-10-21
Fix for autoreply when timezone offset is - from UTC instead of +

hostdrive:
Current autoreply code assumes the timezone offset from UTC always has a + character in front of it. This does not work for timezones in the western hemisphere as the offset has a - in front, and results in an error from web interface.

Instead split date/time at 6th character from right, so it works regardless of + or - TZ offset.

### DIFF
--- a/modoboa/autoreply/models.py
+++ b/modoboa/autoreply/models.py
@@ -1,5 +1,7 @@
 """Postfix autoreply models."""
 
+import re
+
 from django.db import models
 from django.utils import timezone
 from django.utils.formats import localize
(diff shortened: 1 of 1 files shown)


tonioo:
@hostdrive Thanks for reporting this. Can't you use a regexp instead? (using re.split() to split both by + and - characters)

hostdrive:
> @hostdrive Thanks for reporting this. Can't you use a regexp instead? (using re.split() to split both by + and - characters)

Sure I can do that, I reworked my patch. It's a little trickier because the date portion of the string also has - characters and re.split() wants to break the date into multiple pieces. So I joined them back together afterwards. I don't have much Python experience and there might be a better way of doing this.

tonioo:
@hostdrive I've finally created another PR for this (which also includes a frontend fix). I'll close this one but thanks a lot for your contribution !

PR #3801, opened 2025-11-19
Fix #3800 by removing duplicated line

### LINKED ISSUE #3800: DKIM entry duplicated line
Currently the DKIM record shown after clicking "SHOW KEY" contains an additional line with "}})" at the end making it invalid for direct use.

This error comes from https://github.com/modoboa/modoboa/blob/e6269658523c79d77bb929be8d99f952e25c8f43/frontend/src/components/admin/domains/DKIMKeyViewer.vue#L6 as this line duplicates "}})" at https://github.com/modoboa/modoboa/blob/e6269658523c79d77bb929be8d99f952e25c8f43/frontend/src/components/admin/domains/DKIMKeyViewer.vue#L7

pionsys-mhs:
Fix for #3800


tonioo:
@pionsys-mhs Good catch, thanks

PR #3843, opened 2025-12-19
Fix inconsistent autoconfig encryption settings

ntninja:
New Microsoft AutoDiscover and Apple Mail MDM support used hard-coded transport encryption settings (immediate TLS IMAP and, inconsistently, StartTLS SMTP for Microsoft and immediate TLS for Apple). In addition, there was no documentation on the meaning of these values.

Current behavior before PR:
Works if you’re lucky that your settings match.

Can be fixed (and hopefully also understood by admin!) if wrong.

### DIFF
--- a/doc/upgrade.rst
+++ b/doc/upgrade.rst
@@ -181,17 +181,32 @@ Required changes to :file:`settings.py`
      'autoconfig.<MY DOMAIN>',
      'autodiscover.<MY DOMAIN>',
 
-- Add the following variable:
+- Add and adjust the following variable:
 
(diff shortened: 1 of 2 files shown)


tonioo:
@ntninja Thanks for this contribution.
By the way, I'm not sure replacing placeholders is really useful. I have this other PR opened regarding autoconfig service: https://github.com/modoboa/modoboa/pull/3805.
I'll probably merge the two of them into a single if you don't mind.

tonioo:
See https://github.com/modoboa/modoboa/pull/3883

PR #3849, opened 2025-12-27
Allow `bcrypt` 5.0.0 again

ntninja:
The only breaking change is that passwords will no longer automatically be truncated to 72 bytes, so truncate them on input instead.

This reverts commit a7ba1570814e659d0a50faefaf46c5054c0e09f3.


ntninja:
Looking at the trace again after writing a patch for `passlib` to fix this at the correct layer, I realized this is not fixable in Modoboa itself unless we bypass `passlib` entirely.

See #3850 for further discussion!

Issue #2336, opened 2021-07-31
Webmail does not list mails when FTS-Solr is enabled


zsoltbarat:
# Impacted versions

* OS Type: Debian
* OS Version: 9 (stretch)
* Database Type: MySQL
* Database version: 10.1.48-MariaDB-0+deb9u2
* Modoboa: 1.17.0
* installer used: Yes
* Webserver: Nginx / Uwsgi

# Steps to reproduce
1. Configure Solr as FTS engine for Dovecot like:
plugin {
  fts = solr
  fts_autoindex = yes
  fts_solr = break-imap-search url=http://xxxx:8080/solr/
}

2. Launch webmail client in Modoboa

# Current behavior
1. Mails are not listed in overview but are searchable
2. Also quarantine lists mails

Due to how default IMAP search is implemented no mails are found

# Expected behavior
List mails in overview

# Video/Screenshot link (optional)

<img width="1236" alt="grafik" src="https://user-images.githubusercontent.com/15322546/127743140-5afd3717-d139-4209-ab33-f91b65cc1993.png">

zsoltbarat:
dovecot version: 2.2.27

zsoltbarat:
A quick fix would be:
def parse_search_parameters(self, criterion, pattern):
        """Parse search information and apply them."""

        def or_criterion(old, c):
            if old == "":
                return c
            return "OR (%s) (%s)" % (old, c)

        if criterion == u"both":
            criterion = u"from_addr, subject"
        criterions = ""
        for c in criterion.split(','):
            if c == "from_addr":
                key = "FROM"
            elif c == "subject":
                key = "SUBJECT"
            else:
                continue
            criterions = or_criterion(
                criterions, '(%s "%s")' % (key, pattern))

        **if pattern == "":
            criterions = "ALL"**

        if six.PY3:
            criterions = bytearray(criterions, "utf-8")
        elif isinstance(criterions, six.text_type):
            criterions = criterions.encode("utf-8")
        self.criterions = [criterions]

In imaputils.py

Searching for "" does not looks so right anyway.
There is IMAP search ALL for this.

zsoltbarat:
I can provide a proper patch if needed.

tonioo:
@zsoltbarat It would be great :)

PR #3872, opened 2026-01-07
fix(imap): issue when pattern is empty

kryskool:

Problem appear with some FTS backend

Ref #2336

Current behavior before PR:


tonioo:
@kryskool Could you also update tests?

PR #3883, opened 2026-01-16
Some fixes and improvements around the autoconfig service.

tonioo:
Mostly taken from:
 * https://github.com/modoboa/modoboa/pull/3805/changes
 * https://github.com/modoboa/modoboa/pull/3843

### DIFF
--- a/modoboa/autoconfig/views.py
+++ b/modoboa/autoconfig/views.py
@@ -24,6 +24,10 @@ class ConfigBaseMixin:
 
 
 class AutoConfigView(ConfigBaseMixin, generic.TemplateView):
+    """
+    Format documentation:
+    https://wiki.mozilla.org/Thunderbird:Autoconfiguration:ConfigFileFormat
+    """
 
     http_method_names = ["get"]
     template_name = "autoconfig/autoconfig.xml"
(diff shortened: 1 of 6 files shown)


modoboa/autoconfig/views.py
ntninja: Both AutoDiscover and MobileConfig formats support the exact same placeholders as the AutoConfig format? Have you checked this?

ntninja: I substituted them, since I do not think this is the case.

ntninja: Ok, so… Your now always substituting either the received email address or `%EMAILLOCALPART%@%EMAILDOMAIN%` if we are in Auto*Config* and have no email address. Since Modoboa always uses main email address as username there is no issue here.

*However*, you also sending the value of `connection_settings.imap.HOSTNAME` without modification not just on the Auto*Config* but also on the Auto*Discover* and *MobileConfig* endpoints while documenting that one may use Auto*Config*-style placeholders for that field:

```
#  * HOSTNAME: External hostname that clients should connect to
#     * Supports Thunderbird-style placeholders to substitute values from the
#       email address originally entered into the mail client:
#       %EMAILADDRESS%, %EMAILLOCALPART%, %EMAILDOMAIN%
```

This means that if I write a perfectly valid substitution (according the the docs above) into my `settings.py`:

```
EMAIL_CLIENT_CONNECTION_SETTINGS = {
    'imap': {
        'HOSTNAME': 'mail.%EMAILDOMAIN%',
        …
    },
    …
}
```

… that value (`mail.%EMAILDOMAIN%`) will be returned verbatim by *all* three endpoints with Auto*Discover* and *MobileConfig* clients almost certainly either using that value either as-is or failing to parse that XML entirely.

My patch avoided this by performing the substitution server-side for all endpoints. You then silently rewrote it to break substitution again and attributed that to me… :unamused:

tonioo: Ok, sorry for that. But this is not released yet, I'll fix the documentation.

PR #3955, opened 2026-03-02
Initialize counter if key does not exist

jwinterm:
## Problem

When a Redis key does not exist (e.g. after `FLUSHALL`, Redis restart
without persistence, or for newly created mailboxes/domains before any
mail is sent), `redis.hget()` returns `None`.

The `sent_messages` property in `MessageLimitMixin` passes this `None`
directly into arithmetic (`message_limit - None`), which propagates
to `sent_messages_in_percent` and causes:
```
TypeError: int() argument must be a string, a bytes-like object or a
real number, not 'NoneType'
```

This crashes `/api/v2/accounts/` and `/api/v2/domains/`, making the
admin UI completely unusable (500 error on every page that lists
accounts or domains). Domain admin users get a white screen after login.

## How to reproduce

1. `redis-cli FLUSHALL` (or restart Redis without persistence)
2. Log in to Modoboa admin
3. Navigate to Domains or Identities
4. 500 Internal Server Error

## Fix

Guard for `None` in `MessageLimitMixin.sent_messages` and treat a
missing counter as zero messages sent. This is the correct semantic —
a missing Redis key means no messages have been counted yet.

The fix is in `mixins.py` rather than `policyd/utils.py` because the
mixin has the context to interpret what a missing counter means,
while `get_message_counter` is a low-level utility that should
faithfully report what Redis returns.

### DIFF
--- a/modoboa/policyd/utils.py
+++ b/modoboa/policyd/utils.py
@@ -7,5 +7,8 @@ from . import constants
 
 def get_message_counter(key):
     """Return current counter for given key."""
-    rclient = get_redis_connection()
-    return rclient.hget(constants.REDIS_HASHNAME, key)
+    rclient = get_redis_connection(hget_return_type=None)
+    value = rclient.hget(constants.REDIS_HASHNAME, key)
+    if value is None:
+        return None
+    return int(value)


tonioo:
@jwinterm Sorry for the late reply but do you think you could update the tests to cover this case?

jwinterm:
@tonioo added tests here:
https://github.com/modoboa/modoboa/pull/3955/changes/701370d3fdc1fbc184575bbaa3a2bb62abaa5828

tonioo:
@jwinterm Your tests are failing...

jwinterm:
@tonioo sorry about that, but it actually helped uncover a deeper issue.

The original mixin fix (if counter is None: return 0) was dead code. get_redis_connection() sets int as the HGET response callback, so when a key is missing, int(None) raises TypeError inside rclient.hget() before get_message_counter() even returns. The mixin's None check never executed.

I didn't have a local test environment set up initially. After setting one up and running the tests, both new tests failed with the same TypeError the PR was supposed to fix. The actual fix needed to be in get_message_counter() — bypassing the int callback and handling the None/conversion manually.

Summary of commits:

55e86fe — original mixin "fix"
701370d — added tests for domain and mailbox sent_messages when Redis counter is missing
f23379a — the real fix: handle None in get_message_counter() by skipping the int callback
All tests pass locally now.

PR #3975, opened 2026-03-15
Fix empty Actions dropdown on Domains page

### LINKED ISSUE #3974: Domains page: Actions dropdown in toolbar is empty
## Description

On the Domains page, the "Actions" dropdown button in the toolbar is always visible but contains no menu items when clicked. The dropdown renders an empty `VList`.

This differs from the Identities page, where a similar Actions dropdown only appears after selecting rows and correctly populates with bulk action options.

## Steps to reproduce

1. Log in as SuperAdmin
2. Navigate to Domains page
3. Click the "Actions" dropdown button in the toolbar
4. Observe: dropdown opens but contains 0 items

## Expected behavior

The Actions dropdown should either:
- Contain bulk action items (e.g. enable/disable/delete selected domains), similar to the Identities page
- Or not be rendered at all if there are no bulk actions available

## Environment

- Modoboa version: 2.7.1
- Frontend: shipped with modoboa 2.7.1 pip package
- Browser: Chrome 146
- Django: 5.2.11

## Notes

The per-row three-dot action menu (Administrators/Edit/Delete) works correctly. This issue is only with the toolbar-level "Actions" dropdown.

Looking at the bundled `DomainsView` JS, the Actions dropdown renders a `VList` with `density:"compact"` but no children:

```js
a(oe,{location:"bottom"},{
  activator: ...,
  default:l(()=>[
    a(be,{density:"compact"})  // empty VList, no items
  ]),_:1
})
```

themuddfamily:
## Summary

- The Actions dropdown in the Domains list toolbar was always visible but contained an empty `<v-list>` with no menu items
- Added a `bulk_delete` API endpoint (`POST /api/v2/domains/bulk_delete/`) that accepts `{ids: [...], keep_folder: bool}` for deleting multiple domains in a single request
- Frontend now only shows the Actions dropdown when multiple rows are selected (matching the pattern in `IdentityList.vue`), and calls the new bulk endpoint

## Changes

**Backend: `modoboa/admin/api/v2/viewsets.py`**
- Added `bulk_delete` action to `DomainViewSet`
- Accepts `{ids: [1, 2, 3], keep_folder: false}`
- Validates permissions for each domain before deletion

**Frontend: `frontend/src/api/domains.js`**
- Added `bulkDeleteDomains(ids, keepFolder)` method

**Frontend: `frontend/src/components/admin/domains/DomainList.vue`**
- Added `v-if="selected.length > 1"` to Actions menu (same pattern as IdentityList)
- Added `getActionMenuItems()` and `deleteSelectedDomains()` using the bulk endpoint

Fixes #3974

### DIFF
--- a/frontend/src/api/domains.js
+++ b/frontend/src/api/domains.js
@@ -53,6 +53,12 @@ export default {
   deleteDomain(domainId, data) {
     return repository.post(`/${domainResource}/${domainId}/delete/`, data)
   },
+  bulkDeleteDomains(ids, keepFolder = false) {
+    return repository.post(`/${domainResource}/bulk_delete/`, {
+      ids,
+      keep_folder: keepFolder,
+    })
+  },
   createDomainAlias(data) {
     return repository.post(`/${domainAliasResource}/`, data)
   },
--- a/frontend/src/components/admin/domains/DomainList.vue
+++ b/frontend/src/components/admin/domains/DomainList.vue
@@ -29,7 +29,7 @@
             class="flex-grow-0 w-33 mr-4"
           ></v-text-field>
           <slot name="extraActions" />
-          <v-menu location="bottom">
+          <v-menu v-if="selected.length > 1" location="bottom">
             <template #activator="{ props }">
               <v-btn
                 variant="flat"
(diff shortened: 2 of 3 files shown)


tonioo:
@themuddfamily Thanks for this. Actually, I'm not sure that allowing domain multi-deletion is a good idea but if we do it, I'd prefer to implement a single endpoint in the API (bulk mode).
Do you think you could do that instead?

kryskool:
hi @themuddfamily 

Thanks to implement @tonioo suggestions, can you add test for new bulk_delete API ?

Regards

kryskool:
ping @themuddfamily

PR #4038, opened 2026-05-27
Fixed IDOR vulnerability in change password API endpoint.

tonioo:
A domain administrator could use this endpoint to take over a superadmin account by changing its password.

Credits: https://github.com/geo-chen


kryskool (approved):
Works for me

PR #4048, opened 2026-06-03
chore: update metadata and contributing file

kryskool:


pyproject.toml
```diff
@@ -15,13 +15,16 @@ classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Environment :: Web Environment",
     "Framework :: Django :: 5.2",
+	"Framework :: Django :: 6.0",
```
tonioo: We are not using django 6.0 yet

PR #4056, opened 2026-06-23
fix(calendar): Fixed 2 IDOR issues in calendar component

tonioo:
* One in Shared calendar event routes
* One in Access rule routes


kryskool (approved):
Works for me

PR #4057, opened 2026-06-23
fix(webmail): Fixed 2 security issues in webmail component:

tonioo:
* Path traversal issue with HTML image inlining
* IMAP command injection


kryskool (approved):
works for me

PR #4058, opened 2026-06-23
fix(dmarc): bound decompression of aggregate report archives

tonioo:
The DMARC aggregate report importer fully decompressed gzip/zip attachments into memory before parsing, with no size or member-count limits. A remote sender able to deliver mail to the configured report import address could supply a small compressed attachment that expands to a large payload (decompression/zip bomb), exhausting memory and CPU in the importer process (CWE-409 / CWE-400).

Enforce limits before and during decompression in import_archive() and import_report_from_email():
- reject MIME attachments above MAX_COMPRESSED_SIZE before any decompression (also bounds the magic uncompress probe);
- read gzip and xml streams through _read_bounded(), capping output at MAX_DECOMPRESSED_SIZE without materializing the whole stream;
- for zip archives, enforce MAX_ZIP_MEMBERS, a per-member and total uncompressed size cap (checked against ZipInfo.file_size and the real stream, which cannot be fully trusted);
- broaden attachment error handling to (OSError, ValueError, zipfile.BadZipFile) so rejected archives exit with EX_DATAERR (65).

### DIFF
--- a/modoboa/dmarc/lib.py
+++ b/modoboa/dmarc/lib.py
@@ -16,6 +16,7 @@ from dns import resolver, reversename
 import magic
 import io
 
+from django.conf import settings
 from django.db import transaction
 from django.db.models import Q
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


kryskool (approved):
works for me

PR #4075, opened 2026-07-02
fix(autoconfig): parse real Outlook Autodiscover XML request body

BrunoNyland:
This PR solves an issue where email accounts failed to configure automatically in the official Microsoft Outlook client via the AutoDiscover endpoint.

Modoboa's `AutoDiscoverView` attempted to extract the user's email by reading `request.POST["EmailAddress"]`, assuming the data arrived as standard HTTP form data. However, real Microsoft Outlook clients send SOAP requests using a raw XML payload in the request body. Since Django doesn't process this natively as form data, the request resulted in an HTTP 404 error, preventing automatic account setup in Outlook.

A robust fallback was created: if the traditional POST form does not contain the email key, Django will parse the raw XML body (`xml.etree.ElementTree`) to agnostically find the `<EMailAddress>` node. This allows Outlook to properly read IMAP/SMTP configurations and receive an HTTP 200 Success status. An official unit test for this XML payload behavior has also been added. No settings modifications are needed.

**Steps to reproduce / Prove the bug (Ubuntu CLI):**
You can simulate how a real Outlook client queries the server using `curl`. Run the following command against your Modoboa instance:
```bash
# Simulate an Outlook XML POST request
curl -X POST https://your-modoboa-server.com/autodiscover/autodiscover.xml \
  -H "Content-Type: text/xml" \
  -d '<?xml version="1.0" encoding="utf-8"?>
<Autodiscover xmlns="http://schemas.microsoft.com/exchange/autodiscover/outlook/requestschema/2006">
  <Request>
    <EMailAddress>admin@example.com</EMailAddress>
    <AcceptableResponseSchema>http://schemas.microsoft.com/exchange/autodiscover/outlook/responseschema/2006a</AcceptableResponseSchema>
  </Request>
</Autodiscover>'

# ❌ BEFORE PR: The server returns a 404 Not Found HTML page.
# ✅ AFTER PR: The server returns a 200 OK containing the proper <Autodiscover> XML response.
```

---


kryskool:
Hi @BrunoNyland 

Thanks for this PR, it seem correct for me, but can add 2 additionnals test:

1. send and empty EMailAddress empty element (<EMailAddress></EMailAddress>) -> return 404
2. send a non existing email address (e.g. notexists@test.com) -> return 404

Regards,

BrunoNyland:
Glad to help. I added 2 tests to cover the empty email address scenario:

- **add tests for empty email address in Autoconfig and Autodiscover**

I did not implement the second suggested test (non-existing email address), because it would require adding a mailbox existence lookup in `views.py` to validate that the address actually exists before continuing, something like:

```python
    def get_context_data(self, **kwargs):
        # Read body before touching request.POST: once POST is parsed,
        # Django no longer allows access to the raw body.
        emailaddress = _email_address_from_xml_body(self.request.body)
        if not emailaddress:
            emailaddress = self.request.POST.get("EmailAddress")
        if not emailaddress:
            raise Http404
        local_part, domain = split_address(emailaddress)
        if not Mailbox.objects.filter(address=local_part, domain__name=domain).exists():
            raise Http404
        context = super().get_context_data(**kwargs)
        context.update(self.get_common_context(emailaddress))
        return context
```

I think that change would introduce additional behavior and deserves a separate discussion/issue before being included in this PR.

tonioo (approved):

Issue #3245, opened 2024-05-02
IMAPsync configfile configuration and generation


olaf7:
In any v2 version the IMAPmigration lacks the possibility to configure what needs to be migration resulting in the creation of a near .offlineimparc file which in turn errors the migration process.

I am going to add some screenshots v1 panel and v2 panel and comment below.

![image](https://github.com/modoboa/modoboa/assets/835414/f56317de-891b-4d1a-8b43-de14a6e57441)

![image](https://github.com/modoboa/modoboa/assets/835414/bd3d27ac-d69d-4b9d-8cc9-ae012ec999a6)

![image](https://github.com/modoboa/modoboa/assets/835414/3e08ffd2-5128-426c-9acc-f3ffad83e0fa)

![image](https://github.com/modoboa/modoboa/assets/835414/529e1d9a-bfa0-4689-9053-d28f112dbf94)

Problems:

-  no option to define host, port, domain, credentials (or other required or optional settings)
- mismatch between v1 and v2 panel settings (but this is the case in lots of settings in my instance). Here: enable imapsync, create folders
- when changing and saving  the setting to make imapsync active (enable) the change in v2 is not applied when checking the next time Iogin and check the setting

Resulting .offlineimaprc : 
`
[general]
accounts = 
`

tonioo:
@olaf7 The synchronization issue is pretty strange and I don't reproduce it...
About offlineimap settings, once you have enabled the feature, you need to create at least one provider (gmail.com for example) where you'll be able to fill-in connection parameters and more.

olaf7:
As in editing the almost empty config file provided by hand to suit my needs? Or will I be able to edit and or select anything from Modoboa/GUI once the config file is provided with a provider? A webform to supply details/parameters would be nice.

tonioo:
Have you defined a provider and a migration yet?

olaf7:
Please note: effectively this issue should be split into two issues imho. My bad. 1. mismatch settings v1/v2 and 2. IMAP migration
Also: can you comment on how I can help debug this as clearly something is wrong (as can be seen by screenshots), but you cannot reproduce.

To answer your question regarding the imap_sync issue: I created a .offlineimaprc based on the template, but it fails for a couple of reasons:

1.  IIRC no domain and accounts  should be created in advance. There is no existing domain, so no domain admin. Therefor I need to use Modoboa Administrator. This user has no postbox, so no Dovecot login, so offlineimap fails.
[...]

PR #4080, opened 2026-07-03
Better crypto key management

tonioo:
use authenticated encryption for stored documents and derive per-purpose encryption keys with HKDF


kryskool (approved):

kryskool:
link #3245

PR #4083, opened 2026-07-04
feat(core,admin,webmail): add doveadm HTTP API support

tonioo:
Modoboa can now communicate with a remote Dovecot server through the doveadm HTTP API instead of the local doveadm binary. The mode is controlled by the new DOVECOT_OPERATION_MODE setting ('cmd' by default, 'rest' for the HTTP API) along with DOVEADM_API_URL, DOVEADM_API_KEY and DOVEADM_API_TIMEOUT.

A new access layer (modoboa.lib.dovecot) provides both backends behind a common interface and is now used for mailbox location retrieval, scheduled messages handling and password schemes discovery. In rest mode, filesystem operations (mailbox rename/delete) are disabled and password schemes fall back to DOVECOT_SUPPORTED_SCHEMES.

### DIFF
--- a/modoboa/lib/tests/test_dovecot.py
+++ b/modoboa/lib/tests/test_dovecot.py
@@ -222,6 +222,10 @@ class RestModeSchemesTestCase(TestCase):
         from modoboa.core.constants import DOVEADM_PASS_SCHEME_ALARM
         from modoboa.core.password_hashers.utils import get_dovecot_schemes
 
+        # An alarm may have been opened before this test runs (ie. if
+        # doveadm is not installed): get rid of it since we want to
+        # check that the call below doesn't create one.
+        Alarm.objects.filter(internal_name=DOVEADM_PASS_SCHEME_ALARM).delete()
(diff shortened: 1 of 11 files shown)


kryskool (approved):

modoboa/lib/tests/test_dovecot.py
```diff
@@ -0,0 +1,241 @@
...
+    def test_get_dovecot_schemes_fallback(self):
+        from modoboa.admin.models.alarm import Alarm
+        from modoboa.core.constants import DOVEADM_PASS_SCHEME_ALARM
+        from modoboa.core.password_hashers.utils import get_dovecot_schemes
+
+        schemes, status = get_dovecot_schemes()
+        self.assertEqual(schemes, ["{MD5-CRYPT}", "{PLAIN}"])
+        self.assertEqual(status, 1)
+        # No alarm must be opened in this case
+        self.assertFalse(
```
kryskool, line 233: this test seem to return True on CI

PR #4091, opened 2026-07-09
fix(api): add quota_usage to IdentitySerializer for sortable Quota co…

### LINKED ISSUE #4085: Identities admin list: "Quota" column is not sortable (field missing from /api/v2/identities/ payload)
# Impacted versions
* OS Type: Debian/Ubuntu
* OS Version: 24.04
* Database Type: PostgreSQL
* Modoboa: 2.9.0
* installer used: Yes
* Webserver: Nginx

# Current behavior

# Root cause (investigated)

# Expected behavior

# Suggested fix

AdarshJ173:
The Identities admin list (/admin/identities) shows a Quota column but clicking it has no effect. Client-side sorting fails because the API response includes no quota_usage data.                  
                                                                                                                                                                                                        
Fixes #4085                                                                                                                                                                                         
                                                                                                                                                                                                        
Current behavior before PR:                                                                                                                                                                         
                                                                                                                                                                                                        
Quota column header is not clickable — all other columns sort correctly.                                                                                                                            
                                                                                                                                                                                                        
Clicking the Quota column sorts accounts by quota usage percentage. Aliases (which have no mailbox) show "No mailbox" and sort consistently at one end.                                             
                                                                                                                                                                                                        
What Changed                                                                                                                                                                                        
                                                                                                                                                                                                        
- modoboa/admin/api/v2/serializers.py — Added IdentityMailboxSerializer (read-only nested serializer with quota, quota_usage, message_limit, sent_messages_in_percent fields). Added mailbox and quota_usage fields to IdentitySerializer via SerializerMethodField. For user accounts with a mailbox, the payload now includes the nested mailbox object and a flat quota_usage integer. Aliases return null for both.                                                                                                                                                                                             
                                                                                                                                                                                                        
Verification                                                                                                                                                                                        
                                                                                                                                                                                                        
- Syntax verified: python -m py_compile modoboa/admin/api/v2/serializers.py passes.                                                                                                                 
- Full test suite requires PostgreSQL infrastructure not available in this environment. The change is additive (new SerializerMethodField + nested serializer) — existing tests continue to pass because no existing response contracts are modified.                                                                                                                                                       
                                                                                                                                                                                                        
Risks                                                                                                                                                                                               
                                                                                                                                                                                                        
Low risk. The change is additive — new fields on the identity list response. Existing API consumers continue to receive the same fields they always did. The new mailbox field is null for aliases and returns mailbox data for accounts with a mailbox.

### DIFF
--- a/modoboa/admin/api/v2/serializers.py
+++ b/modoboa/admin/api/v2/serializers.py
@@ -5,6 +5,7 @@ import os
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist
 from django.core import validators as dj_validators
 from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext as _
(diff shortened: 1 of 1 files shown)


kryskool:
HI @AdarshJ173 

Thanks for this PR, Can you add tests for this part ?

Regards,

tonioo:
@AdarshJ173 You did not modify the right API endpoint and serializer. This table uses the /accounts endpoint

tonioo:
This fix should be better: https://github.com/modoboa/modoboa/pull/4093

PR #4095, opened 2026-07-10
Expose quota usage in the identities API so the Quota column sorts

### LINKED ISSUE #4085: Identities admin list: "Quota" column is not sortable (field missing from /api/v2/identities/ payload)

TemRevil:
Fixes #4085

On `/admin/identities` every column header sorted the list except **Quota** — the frontend keys that column on `quota_usage` (`IdentityList.vue`) and its cells read `item.mailbox.quota_usage`, but `IdentitySerializer` returned neither field, so the client-side comparator had nothing to order by and the click was a no-op.

### Changes

- **`lib.annotate_quota_usage()`** — extracted the quota-usage annotation (Quota-bytes subquery + percent `Case`) that lived in `AccountViewSet.get_queryset()` and reused it in `lib.get_identities()`, which now also `select_related("mailbox", "mailbox__domain")`. Identity rows get their percentage without per-row `Quota` queries. `AccountViewSet` calls the same helper, so accounts behavior is unchanged.
- **`IdentitySerializer`** now exposes:
  - `quota_usage` at the top level — the frontend's sort key;
  - `mailbox` as `{quota, quota_usage}` — what the Quota cells render (`quota` serialized as a string, same as `MailboxSerializer`, so the template's `quota !== '0'` unlimited check works).
  - Aliases and mailbox-less accounts serialize both as `None`, so they group together when sorting — the behavior the issue asks for.
- **Test**: `IdentityViewSetTestCase.test_list_quota_fields` — sets 5MB used of a 10MB quota, asserts the account row exposes `quota_usage == 50` and `mailbox == {"quota": "10", "quota_usage": 50}`, and that alias rows expose `None` for both.

The serializer falls back to `Mailbox.get_quota_in_percent()` when handed unannotated instances, mirroring the account serializer's fallback.

### DIFF
--- a/modoboa/admin/api/v2/serializers.py
+++ b/modoboa/admin/api/v2/serializers.py
@@ -424,7 +424,41 @@ class IdentitySerializer(serializers.Serializer):
     name_or_rcpt = serializers.CharField()
     tags = TagSerializer(many=True)
     enabled = serializers.BooleanField()
+    quota_usage = serializers.SerializerMethodField()
+    mailbox = serializers.SerializerMethodField()
     possible_actions = serializers.SerializerMethodField()
+
+    def _get_account_mailbox(self, identity):
+        """Return the account's mailbox, None for aliases or mailbox-less accounts."""
+        if not isinstance(identity, core_models.User):
+            return None
+        return getattr(identity, "mailbox", None)
+
+    def get_quota_usage(self, identity) -> int | None:
+        """Quota usage in percent, None for identities without a mailbox.
+
+        Exposed at the top level because the frontend Quota column sorts on
+        this key (aliases group together through the None value).
+        """
+        mailbox = self._get_account_mailbox(identity)
+        if mailbox is None:
+            return None
+        # Reuse the quota_usage annotation set by lib.get_identities() to
+        # avoid a per-row Quota query. Falls back to the model method when
+        # unannotated.
+        if hasattr(identity, "quota_usage"):
+            return int(identity.quota_usage or 0)
+        return mailbox.get_quota_in_percent()
+
+    def get_mailbox(self, identity) -> dict | None:
+        """Quota display data (types mirror MailboxSerializer: quota is a string)."""
+        mailbox = self._get_account_mailbox(identity)
+        if mailbox is None:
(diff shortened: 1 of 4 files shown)


TemRevil:
Ran the two touched suites (IdentityViewSetTestCase and AccountViewSetTestCase) with DB=sqlite in a python:3.13 container with redis available: 37 tests, all passing. That covers the new test_list_quota_fields plus the existing quota_usage ordering tests on the accounts endpoint, so moving the annotation into lib.annotate_quota_usage leaves accounts behavior unchanged.

kryskool:
Hi @TemRevil 

The quota issue have been fix by @tonioo this morning in the [PR4093](https://github.com/modoboa/modoboa/pull/4093)

Can you check if it resolve your issue ?

tonioo:
Already fixed.

TemRevil:
Thanks for the pointer @kryskool. I checked #4093 before opening this one: it enables ordering by quota_usage on the accounts endpoint (/api/v2/accounts/), and this PR actually builds on top of it, reusing its annotation through the shared helper.

Issue #4085 is about a different page though. The identities list (/admin/identities) is served by IdentityViewSet.list() via IdentitySerializer, and that payload still contains no quota_usage or mailbox field after #4093, so clicking the Quota column header there still does nothing. This PR adds those fields to the identities payload, which is what makes that column sortable.

Happy to rebase if anything conflicts, but I believe the issue is still unresolved without it. Could you reopen?

tonioo:
@TemRevil Actually, this endpoint is not used anymore by the frontend because of bad performance (no pagination). I guess it would be better to simply remove it.

TemRevil:
Ah, that explains it — no point sortable-fixing a payload the frontend doesn't fetch anymore. Fine by me to leave this closed and let removal happen separately. Thanks for looking into it.

PR #4098, opened 2026-07-14
fix(api): expose identity quota usage

### LINKED ISSUE #4085: Identities admin list: "Quota" column is not sortable (field missing from /api/v2/identities/ payload)

hasnainkhatri87:
﻿## Summary

Expose mailbox quota usage in the v2 identities API so the Identities page can sort accounts by quota usage.

## Problem

/api/v2/identities/ did not include quota_usage, so the frontend Quota column had no value to sort. Aliases do not have mailbox quotas.

## Implementation

- Add quota_usage to IdentitySerializer.
- Return the mailbox usage percentage for account identities.
- Return 0 for aliases and accounts without a mailbox.
- Add API regression coverage for an account at 50% usage and an alias at 0%.

## Testing

- .\.venv\Scripts\python.exe -m py_compile modoboa\admin\api\v2\serializers.py modoboa\admin\api\v2\tests.py — passed.
- Ad-hoc serializer branch verification — passed.
- .\.venv\Scripts\python.exe tests.py modoboa.admin.api.v2.tests.IdentityViewSetTestCase — blocked on Windows because Modoboa imports the POSIX-only pwd module.
- Full dependency setup was blocked by rrdtool-bindings requiring Microsoft Visual C++ 14+.

Closes #4085

### DIFF
--- a/modoboa/admin/api/v2/serializers.py
+++ b/modoboa/admin/api/v2/serializers.py
@@ -424,7 +424,17 @@ class IdentitySerializer(serializers.Serializer):
     name_or_rcpt = serializers.CharField()
     tags = TagSerializer(many=True)
     enabled = serializers.BooleanField()
+    quota_usage = serializers.SerializerMethodField()
     possible_actions = serializers.SerializerMethodField()
+
+    def get_quota_usage(self, identity) -> int:
+        """Return mailbox quota usage for accounts, or zero for aliases."""
+        if not isinstance(identity, core_models.User):
+            return 0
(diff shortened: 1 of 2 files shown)


tonioo:
Already fixed. See #4085

PR #4100, opened 2026-07-15
security(imap_migration): Fix race condition between time of file creation and restricting file permission bits

### LINKED ISSUE #3004: imap-migration doesn't work properly.
# Impacted versions

* OS Type: Debian/Ubuntu, Centos/Redhat, Arch, Gentoo, Slack, ...
* OS Version: Number or Name
* Database Type: PostgreSQL / MySQL
* Database version: X.y
* Modoboa: 2.1.2
* installer used: Yes ?
* Webserver: Nginx/Apache

# Steps to reproduce

# Current behavior

# Expected behavior

### LINKED ISSUE #3245: IMAPsync configfile configuration and generation

ntninja:
Fixes race condition in between writing IMAP migration configuration file and restricting its permissions. Also adds ability to *not* restrict the file permissions. (Opt-in! I actually have a similar patch to do this for DKIM keys, but didn’t finish/submit it yet.)

Adds tests for both modes, but no regression test as it is very hard to test race conditions reliably without adding a lot of overhead and wasting a lot of dev time.

Please check the remaining Modoboa codebase for similar issues, I’m sure there are some.

### DIFF
--- a/modoboa/imap_migration/management/commands/generate_offlineimap_config.py
+++ b/modoboa/imap_migration/management/commands/generate_offlineimap_config.py
@@ -24,6 +24,12 @@ class Command(BaseCommand):
             default="/tmp/offlineimap.conf",
             help="Path of the generated file",
         )
+        parser.add_argument(
+            "--no-restrict",
+            dest="restrict",
+            action="store_false",
+            help="Do not restrict file permissions on generated OfflineIMAP configuration file",
+        )
 
     def handle(self, *args, **options):
         """Entry point."""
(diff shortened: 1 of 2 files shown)


ntninja:
Also, *any* documentation on how to install IMAP migration would be nice. Although the code makes make think I’m supposed to create a cron job that calls `manage.py generate_offlineimap_config && offlineimap`, is that correct?

Codecov is only complaining about “Indirect Changes” that I didn’t in any way touch.

ntninja:
Fixed a subtle undocumented difference in specifying `opener` as `os.open`, rather than `None`: When using `os.open` the file will be created using umask’d 0o777 permissions, using `None` will instead request umask’d 0o666. Unrestricted default is therefor now `None` since we *don’t* want the file to be created with executable permissions.

tonioo:
@ntninja Could you give more context please? I don't see the issue in the original code on a Linux system

ntninja:
If you create a file with default permissions and *then* restrict its permissions there is a short time window in which the file is not protected and could be opened by an unrelated process before you restrict it.

The new code (optionally with 2nd commit) *creates* the file *restricted* and hence there is no such timeframe where the file *isn’t* protected.

tonioo (approved):

PR #4074, opened 2026-07-01
fix(autoconfig): Thunderbird Mobile Autoconfig changes

pionsys-mhs:
It seems there has been some kind of change since #3805 / #3883 as Thunderbird mobile does not recognise the autoconfig anymore.

The above PR fixes this issue by using the correct placeholder and extracting the mail address domain from the URL used in the request if the not found otherwise.

This fixed it for my servers.

PR #4101, opened 2026-07-15
Fix Thunderbird placeholder support in AutoDiscover/MobileConfig again

ntninja:
Since you immediately broke my code after merging #3883, here’s a fixed version that re-adds support for Thunderbird placeholders. This version only substitutes the placeholders for AutoDiscover/MobileConfig (which always known the email address), keeping them as-is for AutoConfig and hence re-adds the ability to set the `HOSTNAME` substitution to `mail.%EMAILDOMAIN%` (or similar).


kryskool:
@pionsys-mhs can i ask you to check if this PR doesn't broke something since your last PR

Regards,

modoboa/autoconfig/views.py
```diff
@@ -100,11 +121,13 @@ def get(self, request, *args, **kwargs):
...
         local_part, domain = split_address(emailaddress)
+        # Perform server-side placeholder substition, since
+        # MobileConfig clients will not understand
+        # Thunderbird placeholders
+        conn_settings = gen_subst_conn_settings(emailaddress, local_part, domain)
         parts = domain.split(".")
         parts.reverse()
         reverse_domain = ".".join(parts)
-        imap_settings = settings.EMAIL_CLIENT_CONNECTION_SETTINGS["imap"]
-        smtp_settings = settings.EMAIL_CLIENT_CONNECTION_SETTINGS["smtp"]
```
pionsys-mhs, line 107: Why remove this lines instead of changing them to the following?
```
imap_settings = conn_settings["imap"]
smtp_settings = conn_settings["smtp"]
```

This would require less changes in the lines after.

modoboa/autoconfig/views.py
```diff
@@ -11,6 +12,19 @@
 from modoboa.lib.email_utils import split_address
 
 
+def gen_subst_conn_settings(emailaddress: str, local_part: str, domain: str):
```
pionsys-mhs, line 15: I would move this function into `ConfigBaseMixin`. This would also change every call of `gen_subst_conn_settings` to `self.gen_subst_conn_settings`

pionsys-mhs:
@kryskool 

I have checked the modifications from this PR against the recent changes from my PR in #4074.
I manually merged the changes from this PR into one of my development deployments and according to my tests, autoconfig with Thunderbird Mobile and Thunderbird Desktop still works, so nothing is broken.

However, I think this PR modifies a few more lines than absolutely necessary. I left some comments in the code changes.

modoboa/autoconfig/views.py
ntninja: I put it there because `_email_address_from_xml_body` is also global. Probably @tonioo should say which style is preferred…

modoboa/autoconfig/views.py
ntninja: I wanted to avoid introducing even more variables, having all of `conn_settings`, `imap_settings` and `smtp_settings` for essentially the same content felt too much. Again this is a style question @tonioo?

ntninja (commented):
These are both style questions. I commented on the rationale but a project lead should answer what is preferred.

modoboa/autoconfig/views.py
tonioo: I think it's OK to leave it as a function since it would not use anything from ConfigBaseMixin class anway

modoboa/autoconfig/views.py
tonioo: Using intermediate variables was just a matter of readability but it is still readable in this way so fine with me

tonioo (approved):

tonioo:
@ntninja Can you add tests please?

PR #4104, opened 2026-07-15
Add ability to specify post-task command-line to execute after finishing mail log parsing and allow passing `call_command` arguments through RQ job handlers

ntninja:
Primarily, this adds ability to specify post-task command-line to execute after finishing mail log parsing.

On `modoboa-nixos` this is used to rotate and delete the Postfix log file after parsing. NixOS defaults to having no log file, so we also consider it a transient institution as well.

To make this useful all the RQ job handler functions are updated to forward their arguments to `call_command` so that you can customize their command invocations in `cron_config.py`.

### DIFF
--- a/doc-ng/installation/modoboa.md
+++ b/doc-ng/installation/modoboa.md
@@ -309,6 +309,46 @@ numprocs=1
 stopsignal=TERM
 ```
 
+The cron scheduler is *configurable*: During [deployment](#deployment) the file
+`<path to Modoboa instance>/instance/cron_config.py`
+is created containing all currently configured jobs in the form
+`register(<job name>, queue_name="<worker name>", cron="<interval spec>")`.
+
+Editing `cron_config.py` is an advanced feature but may be useful in some
+special cases:
+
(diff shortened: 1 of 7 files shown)


tonioo:
@ntninja Could you add some documentation too on how to use that feature?

ntninja:
@tonioo: I’ve added documentation on the format of `cron_config.py`, it’s basic features used by Modoboa and the how and when to use the `post_cmd` argument for the mail log parser.

I’d also like to point out that I used to implement this by just adding `&& "<path to Postfix>/bin/postfix logrotate && rm /var/log/mail.log.bak` in the crontab. Currently `modoboa-nixos` has a patch to just add the call directly into the `logrotate` implemenation in the Modboa source code, this (perhaps strange-looking) request didn’t come out of nothing, but I don’t think there is any need for similar `post_cmd` options for any other tasks unless someone specifically asks for them. 🙂

ntninja:
(Also moved “potential high system load” hint to the cron scheduler section, since the thing it relates to would now be documented again.)

tonioo (approved):

PR #4105, opened 2026-07-16
feat(admin): Allow generating DKIM keys without restricted permissions

ntninja:
Used on NixOS were permissions are instead restricted on the surrounding directory having only user and group path permissions and the directory’s owning group being that of `rspamd`.

Kinda the inverse of #4100 for DKIM keys.

### DIFF
--- a/modoboa/admin/management/commands/subcommands/_manage_dkim_keys.py
+++ b/modoboa/admin/management/commands/subcommands/_manage_dkim_keys.py
@@ -17,7 +17,7 @@ from ....constants import DKIM_WRITE_ERROR, ALARM_OPENED, DKIM_ERROR
 class ManageDKIMKeys(BaseCommand):
     """Command class."""
 
-    def create_new_dkim_key(self, domain):
+    def create_new_dkim_key(self, domain: str, *, restrict: bool = True) -> None:
         """Create a new DKIM key."""
         storage_dir = param_tools.get_global_parameter("dkim_keys_storage_dir")
         pkey_path = os.path.join(storage_dir, f"{domain.name}.pem")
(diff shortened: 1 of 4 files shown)


modoboa/admin/management/commands/subcommands/_manage_dkim_keys.py
```diff
@@ -17,7 +17,7 @@
 class ManageDKIMKeys(BaseCommand):
     """Command class."""
 
-    def create_new_dkim_key(self, domain):
+    def create_new_dkim_key(self, domain: str, *, restrict: bool) -> None:
```
tonioo: Why do you add a * argument here? Do you expect more arguments?

tonioo:
@ntninja Your test is failing

modoboa/admin/management/commands/subcommands/_manage_dkim_keys.py
ntninja: No, it is just stylistic since it forces the caller to use `, restrict=True)` rather than just `, True)` which IMHO is more readable for boolean in particular. (`*, …` means “everything is a keyword-only argument, while `*args, …` means ”collect remaining positional arguments *and* require following arguments be keyword-only arguments”, if that’s the confusion.)

tonioo: So I guess it is the same than:
```python
def create_new_dkim_key(self, domain: str, restrict: bool = True) -> None:
```
?

ntninja: Almost, that definition can be called as either:

```py
self.create_new_dkim_key("pouet.com", restrict=True)  # OR
self.create_new_dkim_key("pouet.com", True)
```

The definition with star can only be called as:

```py
self.create_new_dkim_key("pouet.com", restrict=True)
```

tonioo: Ok, but I think you're mixing concepts here. Keyword args are meant to be optional as far as I know... And if you want to keep the original behavior, default value should be set to True.

ntninja: Yes, true. They don’t have to be but it usual and makes sense, I just didn’t think about it.

Issue #4102, opened 2026-07-15
Accessing /api/schema-v2/swagger/ fails again


ntninja:
# Impacted versions

* OS Type: NixOS (but also https://demo.modoboa.org/api/schema-v2/swagger/)
* OS Version: 26.05
* Database Type: PostgreSQL
* Database version: 17.10
* Modoboa: 2.9.1
* installer used: N/A
* Webserver: Caddy

# Steps to reproduce

Appears to affect any installation.

# Current behavior

Access /api/schema-v2/swagger crashes with:

```Internal Server Error: /api/schema-v2/swagger/
Traceback (most recent call last):
  File "/nix/store/8mm6msdkryv1kxp10j8is1wk2dkx7bi1-python3-3.13.13-env/lib/python3.13/site-packages/rest_framework/request.py", line 74, in wrap_attributeerrors
    yield
  File "/nix/store/8mm6msdkryv1kxp10j8is1wk2dkx7bi1-python3-3.13.13-env/lib/python3.13/site-packages/rest_framework/request.py", line 232, in user
    self._authenticate()
    ~~~~~~~~~~~~~~~~~~^^
  File "/nix/store/8mm6msdkryv1kxp10j8is1wk2dkx7bi1-python3-3.13.13-env/lib/python3.13/site-packages/rest_framework/request.py", line 385, in _authenticate
    user_auth_tuple = authenticator.authenticate(self)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'AllowAny' object has no attribute 'authenticate'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/nix/store/8mm6msdkryv1kxp10j8is1wk2dkx7bi1-python3-3.13.13-env/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/nix/store/8mm6msdkryv1kxp10j8is1wk2dkx7bi1-python3-3.13.13-env/lib/python3.13/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/nix/store/8mm6msdkryv1kxp10j8is1wk2dkx7bi1-python3-3.13.13-env/lib/python3.13/contextlib.py", line 85, in inner
    return func(*args, **kwds)
[...]

Issue #3924, opened 2026-02-14
401 Unauthorized https://demo.modoboa.org/api/schema-v2/swagger/


nadihighlander:
# Impacted versions

* Modoboa: 2.7.0
* installer used: Yes
* Webserver: Nginx

# Steps to reproduce
Go to this address  https://demo.modoboa.org/api/schema-v2/swagger/

# Current behavior
After upgrading to 2.7.0, API page shows 401 Unauthorized 

# Expected behavior
API page shows swagger api page at address /api/schema-v2/swagger/

# Video/Screenshot link (optional)

nadihighlander:
Even when we feed a valid authorization token in the header, the error is 401 ,

<img width="1341" height="610" alt="Image" src="https://github.com/user-attachments/assets/ab22f71e-145c-4e34-82df-bcbfc1f2b5f7" />

kryskool:
Confirm on https://demo.modoboa.org/api/schema-v2/swagger/ (HTTP 401) logged as admin

tonioo:
@nadihighlander You can update the settings.py file as follows:

<img width="611" height="119" alt="Image" src="https://github.com/user-attachments/assets/93395903-3105-406c-9c1b-78e83474adcc" />

nadihighlander:
> [@nadihighlander](https://github.com/nadihighlander) You can update the settings.py file as follows:
> 
[...]

PR #4107, opened 2026-07-17
fix(core): use authentication classes in SERVE_AUTHENTICATION to prevent swagger crash (refs #4102)

AndrianBalanescu:

Accessing `/api/schema-v2/swagger/` fails with `AttributeError: 'AllowAny' object has no attribute 'authenticate'` (issue #4102).

Current behavior before PR:

Commit e39faf42 (fixing #3924) set `SERVE_AUTHENTICATION` to `['rest_framework.permissions.AllowAny']` in `SPECTACULAR_SETTINGS`. But `SERVE_AUTHENTICATION` expects **authentication** class names, not **permission** class names. When drf-spectacular renders the swagger view, it iterates over the listed classes and calls `.authenticate()` on each — `AllowAny` is a permission class with no `.authenticate()` method, so it crashes.

`SERVE_AUTHENTICATION` uses proper authentication classes (`BasicAuthentication` and `SessionAuthentication`). These allow anonymous access (they simply don't set `request.user` when no credentials are provided), so the swagger page remains publicly accessible — fixing the original #3924 401 issue — without the crash.

Changes:
- `modoboa/core/commands/templates/settings.py.tpl`: replaced `AllowAny` with `BasicAuthentication` + `SessionAuthentication`
- `doc-ng/installation/upgrade.md`: updated the 2.8.0 upgrade instructions to match
- `test_project/test_project/settings.py`: applied the same fix to the test project
- `modoboa/core/api/v2/tests.py`: added `SchemaDocumentationTestCase` with two regression tests that GET the swagger and schema JSON endpoints to verify they return 200


kryskool:
Hi @AndrianBalanescu 

can you remove the .djcache file, and add a rule in gitignore 

Thanks,

kryskool (changes requested):
remove djcache file

AndrianBalanescu:
Removed the `.djcache` files and added `test_project/.gitignore` to keep the cache dir out of version control. Thanks for catching that.

AndrianBalanescu (commented):
already cleaned out in 0fb38cd — those got in through an auto-sync, sorry about that

tonioo:
@AndrianBalanescu Actually, the right fix here would be to use SERVE_PERMISSIONS setting instead of SERVE_AUTHENTICATION (it was a mistake).

tonioo (approved):

Issue #4123, opened 2026-07-29
WebAuthn` /api/v2/fido/registration/begin/` endpoint returns escaped binary username


ntninja:
# Impacted versions

* OS Type: NixOS
* OS Version: 26.05
* Database Type: PostgreSQL
* Database version: 18.0
* Modoboa: 2.9.2
* installer used: No
* Webserver: Caddy

# Steps to reproduce

1. Initiate registering WebAuthn login method

# Current behavior

The client will send a request to /api/v2/fido/registration/begin/ with the user’s session Cookie to retrieve the WebAuthn options set, but the option set’s `.publicKey.user.name` attribute will contain the user’s email address in Python binary string representation with corresponding escaping:
`b'dummy@example.com'`

This is currently irrelevant, but will become a problem if Modoboa starts performing password-less logins, since the stored username in the credential will be of an unexpected format.

Unfortunately, fixing this (removing `.encode("utf-8")` from line 32 of “modoboa/core/fido2_auth.py”) in the server software will *not* update existing credentials stored on clients anywhere. I’m not sure if it is possible to fix the stored username using browser APIs even.

ntninja:
@kryskool @tonioo: This wasn’t fixed. My patch was for fixing another issue, not this one.

kryskool:
HI @ntninja 

I you can, please submit a new PR

Regards

ntninja:
Fixing it going forward is easy, the question what should happen with existing stored credentials?

Updating a credential appears to require something called the WebAuthn Signal API which quite recent and not all browsers or authenticators support it. Even when it *is* available we’d need to compare the entered user name with the server response to and issue an update signal using the API.

When the Signal API is not available *or* unsupported, the only fallback is to interrupt login and ask users to recreate the credential then-and-there, passing the same UID as before so that the original credential ends up being replaced.

So, I can fix the new credential creation, but enabling password-less login will require figuring out a way to do credential updates or have those login fail in non-obvious ways. These same problems also apply when when users change their username…

kryskool:
@tonioo any suggestions from @ntninja comments ?

tonioo:
@ntninja @kryskool This issue needs to be fixed and we can warn users in the release upgrade instructions that they need to recreated credentials...
That being said, I guess the only proper solution for rename cases is to support this new Signal API asap, and warn users that credentials must be recreated when API is not available as @ntninja suggested.

ntninja:
[...]

PR #4124, opened 2026-07-29
fix: Use calling hostname when performing FIDO2 authentication

ntninja:
Calling hostname is validated by Django using `ALLOWED_HOSTS`, so is safe for this operation and it avoids having RP end up being whatever hostname Modoboa happend when the instance was originally created, which may have since changed causing not-very-helpful “Origin and RP-ID do not match” error to be displayed.

Current behavior before PR:
RP-ID is taken from outdated `localconfig.site.domain`, I don’t even *know* to fix that value without database modding.

RP-ID is taken from request.

### DIFF
--- a/modoboa/core/fido2_auth.py
+++ b/modoboa/core/fido2_auth.py
@@ -25,7 +25,7 @@ def get_creds_from_user(user_id: int) -> dict:
 
 
 def begin_registration(request):
-    server = create_fido2_server(request.localconfig.site.domain)
+    server = create_fido2_server(request.get_host())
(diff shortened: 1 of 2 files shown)


ntninja:
Also discovered #4123 while looking into why WebAuthn setup was broken on my instance.

tonioo (approved):

ntninja:
Thank you! Is there any reason to keep `localconfig.site.domain` in config at all? If it will contain stale information anyways, wouldn’t it be better to remove it entirely?

PR #4136, opened 2026-08-13
fix: extract actual token from request.auth instead of str()

### LINKED ISSUE #4131: Webmail 504 timeout: get_imapconnector passes str(request.auth) instead of actual token to Dovecot OAUTHBEARER
## Summary

The webmail's `/api/v2/webmail/mailboxes/` endpoint causes a 504 Gateway Timeout when a user logs in. Dovecot logs show the IMAP authentication hanging for 180 seconds before disconnecting:

```
imap-login: Disconnected: Inactivity during authentication (client didn't finish SASL auth, waited 180 secs): user=<>, method=OAUTHBEARER
```

## Root Cause

In `modoboa/webmail/lib/imaputils.py`, line 851, `get_imapconnector` passes `str(request.auth)` as the token to `IMAPconnector`:

```python
def get_imapconnector(request, **kwargs) -> IMAPconnector:
    return IMAPconnector(request.user.username, str(request.auth), **kwargs)
```

Since **django-oauth-toolkit v3.x**, `AccessToken.__str__()` was changed (for security reasons, to prevent tokens appearing in logs/tracebacks) to return `"AccessToken #<pk>"` instead of the actual token string:

```python
def __str__(self):
    # Never render the token itself: __str__ appears in the admin change
    # page/breadcrumbs, in repr() within tracebacks, and in log output.
    return "AccessToken #{self.pk}".format(self=self)
```

So `IMAPconnector` receives `"AccessToken #42"` as the OAUTHBEARER token. Dovecot's introspection endpoint returns `{"active": false}` for this invalid string, Dovecot sends an error challenge back to the IMAP client, and Modoboa's client (using `_simple_command` which doesn't handle multi-step SASL exchanges) never sends the required `\x01` abort byte — resulting in a 180-second hang per connection attempt.

## Environment

- Modoboa: 5.2.16
- django-oauth-toolkit: 3.4.0

## Fix

Change `str(request.auth)` to `request.auth.token` in `get_imapconnector`:

```python
def get_imapconnector(request, **kwargs) -> IMAPconnector:
    return IMAPconnector(request.user.username, request.auth.token, **kwargs)
```

`request.auth` is always an `AccessToken` instance when the view is reachable (unauthenticated requests return 401 before this code runs), and `AccessToken.token` is the non-nullable CharField holding the actual token string.

This fix has been verified to resolve the issue.

colozhb:
## Fix
Fixes #4131

 passes a string repr of the token object instead of the actual token value, causing OAUTHBEARER auth to hang for 180 seconds.

## Changes
- Extract actual token from  using  or  attributes
- Fall back to  for backward compatibility

### DIFF
--- a/modoboa/webmail/lib/imaputils.py
+++ b/modoboa/webmail/lib/imaputils.py
@@ -933,4 +933,13 @@ def get_imapconnector(request, **kwargs) -> IMAPconnector:
 
     :param request: a ``Request`` object
     """
-    return IMAPconnector(request.user.username, str(request.auth), **kwargs)
+    # Extract the actual token string from the auth object
+    # instead of using str() which returns a useless repr
+    auth = request.auth
+    if hasattr(auth, "token"):
+        token = auth.token
+    elif hasattr(auth, "key"):
+        token = auth.key
+    else:
+        token = str(auth)
+    return IMAPconnector(request.user.username, token, **kwargs)


kryskool:
Hi @colozhb 

It missing the test to reproduce the bug, can you add it

Regards,

tonioo:
@colozhb It would be better to pin django-oauth-toolkit version (to something like '>4') and only keep the appropriate case in your fix.

PR #4137, opened 2026-08-13
feat(webmail): Add IMAP folder subscriptions management

tonioo:
Add a "Subscriptions" entry to the mailbox operations menu that opens a dialog listing the whole folder tree with a checkbox per folder to view and edit the current account's IMAP subscriptions.

The webmail sidebar now only displays subscribed folders (LIST with the SUBSCRIBED return option). Subscription toggling is applied as a batch on confirmation.


kryskool (approved):
Works for me

PR #4138, opened 2026-08-13
fix(webmail): reliable unread counters in the mailbox sidebar

tonioo:
Unread counters were computed with the \Marked/\UnMarked LIST heuristic, which only tracks *recent* activity since the last SELECT. A mailbox holding unseen messages but flagged \UnMarked (e.g. subfolders visited earlier) was skipped, so no counter was displayed.

Backend: compute the unseen count for every selectable mailbox actually returned by LIST (skipping \Noselect ones) instead of relying on that heuristic. The counting is now recursive to also cover deep-loaded trees.

Frontend: use mailbox.unseen as the single source of truth for all folders (current one included), refresh the current mailbox counter in place on mbCounterKey bumps, refresh a mailbox counter when it is selected (a message may have arrived meanwhile), and include the counter refresh in the junk actions and the 5-minute auto-refresh.

Add a regression test covering the \UnMarked and \Noselect cases.

### DIFF
--- a/frontend/src/components/webmail/EmailList.vue
+++ b/frontend/src/components/webmail/EmailList.vue
@@ -369,6 +369,7 @@ const fetchEmails = () => {
 
 const autoRefreshContent = () => {
   fetchEmails()
+  reloadMailboxCounters()
 }
 
 const submitSearch = () => {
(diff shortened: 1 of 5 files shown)


kryskool (approved):
works for me

PR #4149, opened 2026-08-26
fix(admin): dedupe alias recipients on resolved target

tonioo:
An alias recipient using an address extension (foo+bar@dom) is stripped of its extension during resolution, so foo@dom and foo+bar@dom resolve to the same mailbox. add_recipients deduplicated on the raw address only, letting both be inserted and violating the (alias, r_mailbox) unique constraint with an IntegrityError -- notably when saving the pointed account re-syncs its aliases.

Deduplicate on the resolved r_mailbox/r_alias target before inserting. External recipients keep their address-based dedup.


kryskool (approved):
works for me

PR #4151, opened 2026-08-28
fix(core): resolve extension version from installed package metadata

tonioo:
The version displayed for an extension came from the `version` class attribute hardcoded in each plugin's modo_extension.py. When a plugin is released without bumping that attribute, Modoboa keeps advertising the old version (and the update check compares against a stale value).

Resolve the version from the installed distribution metadata, like the core already does, and fall back to the declared attribute for local extensions or source checkouts.

### DIFF
--- a/modoboa/core/extensions.py
+++ b/modoboa/core/extensions.py
@@ -1,10 +1,30 @@
 """Extension management."""
 
+import functools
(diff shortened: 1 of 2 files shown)


kryskool (approved):
works for me

PR #4153, opened 2026-08-28
fix(core): harden SMS password recovery on the Django side

tonioo:
Password recovery is served by the Django views: the Vue3 frontend delegates authentication to OIDC and its recovery screens were removed with the OIDC migration. The rate limiting, however, only existed on the unused API v2 endpoints.

Rate limit the Django views. `ThrottleViewMixin` applies the existing DRF throttle classes to plain class based views: they only need `request.META` and the cache, so both sides share the same scopes, rates and counters and an attacker cannot double a quota by alternating between them. `PasswordResetConfirmView` is subclassed to get the same treatment. A resend consumes the request budget rather than the larger verification one, since it sends a billable SMS, and the verification counter is only cleared on a genuine success so wrong attempts keep costing something.

Remove the API v2 recovery endpoints, which had no consumer left, along with their serializers, exceptions, tests and the orphan frontend calls.

Fix three defects in the remaining views:
- an empty phone number passed the `isnull=False` filter and an SMS was sent to an empty recipient; disabled accounts were eligible too, unlike in the email branch of the same form
- the SMS branch matched the address case sensitively while the email branch used `iexact`
- a code was only accepted for about 90 seconds, which an SMS delivery plus typing can easily exceed. Generation and verification now go through shared helpers and the accepted window is widened to at least 10 minutes
- as a side effect, the resend path passed a `PhoneNumber` object where the provider payload expects a string

### DIFF
--- a/frontend/src/api/auth.js
+++ b/frontend/src/api/auth.js
@@ -4,15 +4,6 @@ export default {
   requestToken(data) {
     return repository.post('/token/', data)
   },
-  recoverPassword(data) {
-    return repository.post('/password_reset/', data)
-  },
-  checkSmsTotp(data) {
-    return repository.post('/sms_totp/', data)
-  },
-  changePassword(data) {
-    return repository.post('/reset_confirm/', data)
-  },
   beginFidoRegistration() {
     return repository.post('/fido/registration/begin/')
   },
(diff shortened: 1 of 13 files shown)


kryskool (commented):
works for me

kryskool (approved):

PR #4154, opened 2026-08-28
feat(admin): add bulk delete action to domain list

### LINKED ISSUE #3974: Domains page: Actions dropdown in toolbar is empty

tonioo:
Add a dedicated POST /api/v2/domains/bulk_delete/ endpoint, mirroring what already exists for accounts.

The queryset is filtered by get_for_admin() so any id the user can't access returns a 404 without deleting anything, and deleting the domain owning one's own mailbox is refused. Deletion goes through Domain.delete() to keep the business logic (access revocation, quotas, auto_account_removal).

DjangoModelPermissions maps POST to the "add" permission, which is wrong for a deletion endpoint, so a CanDeleteDomain permission class is used for this action only: super admins and resellers can call it, domain admins get a 403.

On the frontend side, the (so far empty) "Actions" menu of the domain list now offers a Delete entry when several domains are selected.

fix #3974

### DIFF
--- a/frontend/src/api/domains.js
+++ b/frontend/src/api/domains.js
@@ -53,6 +53,9 @@ export default {
   deleteDomain(domainId, data) {
     return repository.post(`/${domainResource}/${domainId}/delete/`, data)
   },
+  bulkDeleteDomains(data) {
+    return repository.post(`/${domainResource}/bulk_delete/`, data)
+  },
   createDomainAlias(data) {
     return repository.post(`/${domainAliasResource}/`, data)
   },
--- a/frontend/src/components/admin/domains/DomainList.vue
+++ b/frontend/src/components/admin/domains/DomainList.vue
@@ -29,7 +29,10 @@
             class="flex-grow-0 w-33 mr-4"
           ></v-text-field>
           <slot name="extraActions" />
-          <v-menu location="bottom">
+          <v-menu
+            v-if="canDeleteDomain && selected.length > 1"
+            location="bottom"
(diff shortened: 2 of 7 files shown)


kryskool (approved):
works for me

PR #4155, opened 2026-08-28
feat(auth): patch fido registration name

### LINKED ISSUE #4123: WebAuthn` /api/v2/fido/registration/begin/` endpoint returns escaped binary username

ntninja:
Fixes #4123 as best as is possible

Current behavior before PR:
PassKeys are created with `b'<username>'` as user name. User name changes are not propagated to authenticators.

PassKeys are created with correct user names. The WebAuthn Signal API is used to propagate the current user name and user display name on login to compatible authenticators on compatible browsers (not many), also the current list of accepted credentials is signalled since I was at it.

Notably the authoritative user name and display name are only returned by the server *after* solving the 2FA challenge to give maximum assurance that no private user data is leaked even to attackers knowing a user’s username and password.

### DIFF
--- a/modoboa/core/api/v2/tests.py
+++ b/modoboa/core/api/v2/tests.py
@@ -364,9 +364,37 @@ class AuthenticatorData(bytes):
 class FIDOViewSetTestCase(ModoAPITestCase):
     @mock.patch("fido2.server.Fido2Server.register_complete")
     def test_registration(self, register_complete_mock):
+        self.maxDiff = None
+
         url = reverse("v2:fido-registration-begin")
         resp = self.client.post(url)
         self.assertEqual(resp.status_code, 200)
+        self.assertIn("publicKey", resp.json())
+        self.assertIn("challenge", resp.json()["publicKey"])
+        self.assertIn("pubKeyCredParams", resp.json()["publicKey"])
+        self.assertIn("user", resp.json()["publicKey"])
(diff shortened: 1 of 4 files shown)


ntninja:
One further improvement here would be to also signal a “display name” change when changing settings in `/account` (and technically also a ”name” change when admins change their email address in `/admin/identities`)

tonioo (approved):

PR #4160, opened 2026-09-03
perf: speed up modo repair on large instances

tonioo:
The repair command issued at least one query per object, whatever the number of problems to actually fix. On a 500 mailbox instance with a healthy database, a run cost 2537 queries for the ownership check and 1513 for the alias check; both are now constant (5 and 3).

- detect ownerless objects with a single anti-join per model instead of one get_object_owner() call per object
- select_related() the relations fix_owner() dereferences; the Mailbox one was an N+1 triggered by Mailbox.__init__ reading full_address
- look for missing internal aliases and alias recipients by loading what already exists first, instead of two get_or_create() per mailbox
- fetch super users once per queryset and hand them to grant_access_to_object(), which now bulk creates their entries and only writes is_owner when it actually changes
- run each check in a single transaction
- index ObjectAccess on (content_type, object_id): every lookup that ignores the user could not use the unique index, whose leading column is user_id. This mostly benefits get_object_owner(), ungrant_access_to_object() and Domain.admins rather than repair itself (2.457 ms -> 0.008 ms on a 60k rows table with PostgreSQL 16).

Alias recipient unicity is now checked on (alias, r_mailbox), the model unique_together, instead of (alias, address, r_mailbox): an existing row whose address drifted is left alone rather than raising IntegrityError.


kryskool (approved):
Works for me

PR #4161, opened 2026-09-03
perf: speed up super admin creation and promotion

tonioo:
Promoting an account to SuperAdmins grants it access to every object of the instance, through grant_access_to_objects(). That function ran one get_or_create() per object, and was handed full querysets whose model instances were built for their primary key only -- Mailbox.__init__ even queries the database on its own to read full_address.

Load the already granted object ids in a single query, then bulk create what is missing. Querysets are consumed with values_list("pk") so no model instance is built; a plain list of objects is still accepted.

With 414 objects (208 accounts, 204 mailboxes, 2 domains):

  promotion             1874 queries / 198 ms -> 23 queries / 14 ms
  creation              1891 queries          -> 36 queries
  promotion replayed     628 queries          -> 15 queries

The grants are also wrapped in a transaction: a partially promoted super admin would silently lack access to some objects, and ATOMIC_REQUESTS does not cover management commands nor RQ tasks.

### DIFF
--- a/modoboa/admin/handlers.py
+++ b/modoboa/admin/handlers.py
@@ -1,6 +1,7 @@
 """Django signal handlers for admin."""
 
 from django.contrib.contenttypes.models import ContentType
+from django.db import transaction
 from django.db.models import signals
 from django.dispatch import receiver
 from django.utils.translation import gettext as _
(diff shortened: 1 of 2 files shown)


kryskool (approved):
works for me

PR #4162, opened 2026-09-03
fix: stop leaving objects without an owner on role downgrade

tonioo:
Setting a role deletes every ObjectAccess entry of the account when it loses its privileges (a super admin being demoted, or anybody becoming a SimpleUsers), including the is_owner ones. Nothing was reattributed, so each object the account owned was left without any owner until somebody ran "modo repair".

Add permissions.transfer_ownership(), which gives those objects to another account -- the owner of the demoted one, then any other active super admin -- and call it before the accesses are dropped. Existing non owner entries of the new owner are promoted rather than duplicated.

The account deletion handler now shares that helper, which fixes two of its own problems:

* from_user was only assigned when a request was available, but read unconditionally when the account owned itself. Deleting the default admin outside of a request (management command, RQ task) raised UnboundLocalError instead of working.
* it walked ObjectAccess entries one by one and dereferenced the generic foreign key of each, which instantiates the target object. Deleting a super admin of a large instance meant two queries per owned object, and Mailbox.__init__ queries the database on its own.

### DIFF
--- a/modoboa/admin/tests/test_ownership.py
+++ b/modoboa/admin/tests/test_ownership.py
@@ -1,0 +1,133 @@
+"""Tests about object ownership."""
+
+from django.core import management
+
+from modoboa.core.models import User
+from modoboa.lib.permissions import get_object_owner
+from modoboa.lib.tests import ModoTestCase
+from .. import factories, models
+
+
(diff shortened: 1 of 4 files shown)


kryskool (approved):
works for me

PR #4172, opened 2026-09-11
Webmail: sandbox message display and embed inline images

tonioo:
Render message bodies in a sandboxed iframe (srcdoc, no scripts,
opaque origin) with a restrictive CSP, so untrusted HTML can never
reach the application context even if the server-side cleaner is
bypassed. Remote content is only allowed when links are enabled.

Inline images are no longer written to the shared, publicly served
media/webmail directory (UIDs are only unique per mailbox, so images
could leak between users). They are now embedded as data: URIs,
restricted to safe raster image types.

### DIFF
--- a/frontend/src/views/webmail/EmailView.vue
+++ b/frontend/src/views/webmail/EmailView.vue
@@ -242,13 +242,13 @@ const emailDocument = computed(() => {
     "style-src 'unsafe-inline'",
     "form-action 'none'",
   ].join('; ')
-  return (
-    '<!DOCTYPE html><html><head><meta charset="utf-8">' +
-    `<meta http-equiv="Content-Security-Policy" content="${csp}">` +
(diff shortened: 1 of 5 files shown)


frontend/src/views/webmail/EmailView.vue
```diff
@@ -215,8 +227,35 @@ const close = () => {
...
+  const remoteSrc = enableLinks.value ? ' https: http:' : ''
+  const csp = [
+    "default-src 'none'",
+    `img-src data:${remoteSrc}`,
+    `media-src data:${remoteSrc}`,
+    `font-src data:${remoteSrc}`,
+    "style-src 'unsafe-inline'",
+    "form-action 'none'",
+  ].join('; ')
+  return (
```
kryskool: Use only backtick

kryskool (changes requested):

PR #4173, opened 2026-09-11
fix: webmail: escape mailbox names in IMAP commands

tonioo:
The imap4-utf-7 codec leaves double quotes and backslashes untouched, and mailbox names were wrapped in quotes without escaping them. A crafted name (e.g. from the "mailbox" request parameter) could break out of the quoted string and inject extra command arguments.

Add quote_mailbox_name(), which rejects control characters and backslash-escapes quotes and backslashes, and use it for every command taking a mailbox name, including the LIST pattern.


kryskool (commented):
works for me

kryskool (approved):

PR #4175, opened 2026-09-11
fix(webmail): webmail fix reply to forward

tonioo:
- Reply-To was serialized with an EmailField while it is parsed as a
  list of addresses: the API returned the Python list as a string,
  which the compose form then used as the recipient. It is now a list
  of structured addresses and replies target every Reply-To address.
- Forwarding pre-filled the recipient with the original sender; the
  form now leaves recipients empty when forwarding.
- "Reply all" put "Name <address>" values in Cc, which the API rejects
  (EmailField): it now uses bare addresses, without duplicates or
  addresses already in To.


kryskool (approved):
works for me

PR #4176, opened 2026-09-11
fix(webmail): handle SMTP network errors and stuck scheduled messages

tonioo:
- Only SMTPResponseException and SMTPRecipientsRefused were caught: a refused connection, a timeout, a disconnection or a TLS failure returned a 500 on direct sending and crashed the scheduled sending job. All SMTP and network errors (OSError) are now reported, and bytes SMTP replies are decoded instead of shown as "b'...'".
- Failing to store a copy in the Sent folder no longer reports an error for a message that has actually been sent (the user would send it again); the failure is logged.
- The scheduled sending job flags the message as failed on any unexpected error, ignores messages cancelled or no longer in the SENDING state, and the error is truncated to the field length.
- Messages left in the SENDING state for more than an hour (job lost by the queue) are flagged as failed so the user can reschedule them.
- Rescheduling a failed message now puts it back in the SCHEDULED state: it was never sent again before.

### DIFF
--- a/modoboa/webmail/constants.py
+++ b/modoboa/webmail/constants.py
@@ -1,9 +1,9 @@
 """Webmail constants."""
 
+from datetime import timedelta
 from enum import Enum
 
 from django.utils.translation import gettext_lazy as _
-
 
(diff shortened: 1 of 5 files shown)


kryskool (commented):
works for me

PR #4178, opened 2026-09-11
fix(webmail): report refused mailbox operations instead of crashing

tonioo:
- WebmailInternalError applied a text regex to the raw bytes returned by imaplib: a refused rename or delete (e.g. unknown folder) raised a TypeError, hence a 500. The reason is now decoded first, which also removes the "b'...'" shown in create/subscribe errors.
- Folder operations (create, rename, delete, subscribe, unsubscribe) share a helper turning NO responses and imaplib errors (BAD responses) into MailboxOperationError, a 400 carrying the server message: these are user errors, not internal ones.
- Invalid values that cannot be sent to the IMAP server (mailbox name, UID, part number, search pattern) raise InvalidImapArgument, a 400, instead of a 500.
- The real imaplib exception class is kept in IMAP4Error, so catching it keeps working when tests mock imaplib.IMAP4.

### DIFF
--- a/modoboa/webmail/exceptions.py
+++ b/modoboa/webmail/exceptions.py
@@ -11,6 +11,11 @@ class WebmailInternalError(InternalError):
     errorexpr = re.compile(r"\[([^\]]+)\]\s*([^\.]+)")
 
     def __init__(self, reason, ajax=False):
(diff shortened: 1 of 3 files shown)


kryskool (approved):
works for me

PR #4179, opened 2026-09-11
fix(webmail): keep addresses whole when display names contain commas

### LINKED ISSUE #2347: Email address name with "," invalid on replying email
# Impacted versions

* OS Type: Server: Centos/Redhat, Client: Win10, Browser: Firefox
* OS Version: 7
* Database Type: PostgreSQL
* Database version: OOB from the Modoboa installer
* Modoboa: 
* installer used: Yes
* Webserver: Nginx

# Steps to reproduce
I don't know if other Email systems can replicate the issue. Reply an email sent from Outlook by a sender which his/her name has "," in it like "Berhane, Kidane T - Eagan, MN". The "To" email address on the reply email will become three separate addresses, "Berhane", "Kidane T - Eagan", and "MN". Error message "Enter a valid email address." returned when the email sent.  

# Current behavior
If email address name have "," on it, system split the email address into multiple addresses.

# Expected behavior
Email addresses with "," in the names should be handled correctly.

tonioo:
Address lists were split on every comma, so a display name such as "Berhane, Kidane T - Eagan, MN" (common with Outlook) produced several invalid addresses: replying to such a message filled the recipient fields with fragments that failed validation.

- parse_address_list now relies on email.utils.getaddresses, which follows the address syntax (quoted names, folding).
- The message view parses address headers from their raw value: decoding encoded-words first turned an encoded comma (=?utf-8?q?Doe=2C_John?=) into an address separator.

Closes #2347

### DIFF
--- a/modoboa/webmail/lib/imapemail.py
+++ b/modoboa/webmail/lib/imapemail.py
@@ -22,6 +22,9 @@ from . import imapheader
 from .imaputils import get_imapconnector, validate_imap_uid, BodyStructure
 from .utils import decode_payload
 
+# Headers holding addresses, parsed from their raw value
+ADDRESS_HEADERS = ("From", "To", "Cc", "Bcc", "Reply-To")
+
 
 class ImapEmail(Email):
     """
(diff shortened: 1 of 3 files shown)


kryskool (approved):
works for me

PR #4180, opened 2026-09-11
fix(webmail): flag original messages as answered or forwarded

tonioo:
The \Answered and $Forwarded flags were never set: the helpers existed but nothing called them, so replies and forwards never showed up in the message list.

- The compose form sends the mailbox and UID of the message being replied to or forwarded; once the message is sent, the API flags it. The three fields must be provided together. Failing to set the flag does not fail the sending, the message is already gone.
- msg_answered and msg_forwarded passed the UID as a string to a helper joining a list: UID "123" would have flagged messages 1, 2 and 3. They now pass a list of one validated UID.


kryskool (approved):
works for me

PR #4181, opened 2026-09-11
fix(webmail): store attachments privately and clean up resources

tonioo:
Attachments were stored in MEDIA_ROOT/webmail, which the web server may serve publicly, and nothing was ever cleaned up.

- Attachments (messages being written, scheduled messages) now live in WEBMAIL_ATTACHMENTS_ROOT, by default BASE_DIR/webmail_attachments, next to MEDIA_ROOT and never inside it. The storage gives files no URL and stored names can't point outside the directory. A migration moves the files of existing scheduled messages.
- Compose sessions get one Redis key each, expiring after a day without activity, instead of a hash per user growing forever.
- An hourly job deletes files no compose session nor scheduled message refers to anymore; deleting a scheduled attachment (sent or cancelled message) deletes its file; scheduling a message drops its compose session.
- make_body_images_inline no longer embeds files from the attachments directory or the legacy media/webmail one: a user knowing the name of another user's upload could send it to themselves.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ test_project/cert.crt
 test_project/cert.key
 test_project/envfile
 test_project/media
+test_project/webmail_attachments
 
 # Docker
 docker/collections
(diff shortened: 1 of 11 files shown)


kryskool (approved):
works for me

PR #4182, opened 2026-09-12
perf(webmail): use a single IMAP connection per request

tonioo:
Every operation opened its own connection, each costing a TCP connection, a TLS handshake and an authentication: sending a message opened three of them (Sent copy, draft removal, flagging). ImapEmail also opened one in its constructor and only closed it in __del__, so an exception leaked it until garbage collection.

- get_imapconnector returns a connector cached on the request; nested "with" blocks share it and only the first one authenticates.
- Webmail viewsets close that connection when the request ends, even when the view raised.
- The message view no longer fetches BODYSTRUCTURE twice.
- __enter__ tested the load_namespaces method (always true) instead of the with_namespaces flag.


kryskool (approved):
works for me

PR #4183, opened 2026-09-14
perf(webmail): get unseen counters with the mailboxes list

tonioo:
Displaying the folder tree sent one STATUS command per folder, on top of the LIST command that already asked the server for counters and then ignored them.

- LIST now asks for STATUS (MESSAGES UNSEEN) and the counters sent back with the reply are used, so one command brings the whole tree.
- That RETURN option is only asked when the server advertises LIST-STATUS (RFC 5819); it was sent unconditionally before. Servers without it keep the previous behaviour, one STATUS per folder.


kryskool (approved):
works for me

PR #4184, opened 2026-09-14
fix(webmail): really move messages out of the source mailbox

tonioo:
Moving (and deleting, which moves to the trash) copied the messages and only flagged the originals as deleted. They stayed in the source mailbox: hidden by the webmail, which filters them out, but still visible in other mail clients and still counted twice in the user quota until someone compressed the folder by hand.

- The MOVE extension (RFC 6851) is used when the server advertises it: one atomic command, no leftovers.
- Otherwise the copied messages are expunged by UID (UIDPLUS, RFC 4315). A mailbox-wide EXPUNGE is never sent: it would also remove messages another client of the user flagged as deleted.
- Servers supporting neither keep the previous behaviour.
- delete_mail (draft replacement, scheduled copy) expunged the whole folder the same way; it now targets the message only.
- Rescheduling a message whose IMAP copy is unknown no longer fails.


kryskool (approved):
Works for me

PR #4201, opened 2026-09-16
fix(webmail): limit the number and the total size of attachments

tonioo:
The maximum size applied to each file only: a message could carry any number of them, and their total size was unbounded.

Add two global parameters, max_attachments_total_size and max_attachments_count. The total size is enforced by the upload handler, which interrupts the transfer as soon as the message would exceed it, and the number of files is checked before the upload starts.


modoboa/webmail/lib/attachments.py
```diff
@@ -132,6 +132,11 @@ def _create_attachment_file():
         raise InternalError(str(e)) from None
 
 
+def attachments_size(session: dict) -> int:
+    """Return the total size of the attachments of a compose session."""
+    return sum(att.get("size") or 0 for att in session.get("attachments", []))
```
kryskool: better to use 

`att.get("size", 0)`

kryskool (changes requested):

kryskool (approved):

